PR #693, opened 2014-06-04
Lecture capture modules

mdale:
Lecture capture modules


itaykinnrot:
we have the plugin ready no need for demo

PR #1245, opened 2014-12-04
FEC-2402 #comment dont display an error if the error message is already ...

itaykinnrot:
...set #time 2h


amirch1:
@itaykinnrot - do we still want to add this?

PR #1369, opened 2014-12-29
Template enhancements

ranyefet:
This pull request add few enhancements for tempates.

1. Allow loading template files from ```kwidget-ps``` by using ```{html5ps}``` prefix.
2. Support for registering template helpers.

In ```myPlugin.js```:
```javascript
mw.util.registerTemplateHelper('trDateFormat', function( unixTimestamp ){
	var d = new Date(unixTimestamp*1000);
	var months = ['Jan','Feb','Mar','Apr','May','Jun','Jul','Aug','Sep','Oct','Nov','Dec'];
	var year = d.getFullYear();
	var month = months[d.getMonth()];
	var day = d.getDate();
	var hour = d.getHours();
	var min = d.getMinutes();
	var sec = d.getSeconds();
	return day + '-' + month + '-' + year + ' ' + hour + ':' + min;
});
```

In ```myTemplate.html```:
```html
<span>Entry created at: <%=helpers.trDateFormat(entry.createdAt);%></span>
```

Please review & merge

### DIFF
--- a/modules/KalturaSupport/UiConfResult.php
+++ b/modules/KalturaSupport/UiConfResult.php
@@ -18,7 +18,10 @@ class UiConfResult {
 	var $playerConfig = null;
 	var $noCache = null;
 	var $isPlaylist = null;
-	var $isJsonConfig = null;	
+	var $isJsonConfig = null;
+
+	// These properties are handled differently, we don't want to resolve them automatically
+	var $propsToIgnoreWhenResolvingResource = array('templatePath');
 	
 	function __construct( $request, $client, $cache, $logger, $utility ) {
 
@@ -240,6 +243,7 @@ class UiConfResult {
 				$this->walkThroughPluginAttributes( $attrValue , $refrencePathInObject[$attrKey] );
 			}
 			else {
+				if( in_array($attrKey, $this->propsToIgnoreWhenResolvingResource) ) continue;
 			    $refrencePathInObject[$attrKey] = $this->resolveCustomResourceUrl( $attrValue );
 			}
 		}
--- a/modules/KalturaSupport/resources/mw.KBasePlugin.js
+++ b/modules/KalturaSupport/resources/mw.KBasePlugin.js
@@ -79,6 +79,7 @@ mw.KBasePlugin = Class.extend({
 		data.player = this.embedPlayer;
 		data.entry = this.embedPlayer.kalturaPlayerMetaData;
 		data.entryMetadata = this.embedPlayer.kalturaEntryMetaData;
+		data.helpers = mw.util.getTemplateHelpers();
 
 		// First get template from 'template' config
 		var rawHTML = this.getConfig( 'template', true );
--- a/resources/mediawiki/mediawiki.util.tmpl.js
+++ b/resources/mediawiki/mediawiki.util.tmpl.js
@@ -6,6 +6,14 @@
 	"use strict";
 
 	var cache = {};
+	var helpers = {};
+	mw.util.registerTemplateHelper = function( name, callback ){
+		helpers[ name ] = callback;
+	};
+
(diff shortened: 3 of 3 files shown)


ranyefet:
@mdale how do you suggest implementing helpers/filters for templates?
Say, you want to display duration of the video which comes in seconds i.e ```80``` and you want to show it as: ```01:20``` how would you do that?
You need to have a way to register custom filters/format that you could use in templates.
I needed a way to do that, without changing the Playlist plugin code.

Do you have better solution?

mdale:
Both the original example and what you mentioned could be done with the existing [format functions](https://github.com/kaltura/mwEmbed/blob/master/modules/KalturaSupport/resources/mw.KDPMapping.js#L17) something like: 
```html
<span>Entry created at: {mediaProxy.entry.createdAt|dateFormat}</span>
```
for time format we also have a handler: 
```html
<span> Advertisements: {sequenceProxy.timeRemaning|timeformat}</span> 
```
for a core localized template presently a little messy: 
```html
<span><%=gM( 'ks-created-on', %>{mediaProxy.entry.createdAt|dateFormat}<=%);%>></span>
```
could be cleaned up if we had a kaltura specific localization call that ran evaluations on subtitled properties.

It terms of plugins adding formaters; we could introduce something like what you have above for adding new formatFunctions i.e: 
```javascript
mw.KDPMappingAddFormater('trDateFormat', function( unixTimestamp ){
    var d = new Date(unixTimestamp*1000);
    var months = ['Jan','Feb','Mar','Apr','May','Jun','Jul','Aug','Sep','Oct','Nov','Dec'];
    var year = d.getFullYear();
    var month = months[d.getMonth()];
    var day = d.getDate();
    var hour = d.getHours();
    var min = d.getMinutes();
    var sec = d.getSeconds();
    return day + '-' + month + '-' + year + ' ' + hour + ':' + min;
});
```
then {mediaProxy.entry.createdAt|trDateFormat} could be used in templates.

mdale:
P.S there are of course more standard solutions for date formatting and localization that should be used, I imagine the example was a PS style use case for a specific formater; but in core localization would have to be taken into consideration.  
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString

ranyefet:
@mdale 
```<span>Entry created at: {mediaProxy.entry.createdAt|dateFormat}</span>```
the ```dateFormat``` is only available when using evaluated strings.
But in my case, I needed to format the playlist items object which is not exposedvia evaluated string.
So your suggestion will not work.

I do agree that we should have single place to store this helpers and not store them in multiple locations.

**Comment by mdale** [comment]
@ranyefet  what do you mean "wont work" ? we already use these formaters in templates. For example the info template ( that you wrote ) : https://github.com/kaltura/mwEmbed/blob/master/modules/KalturaSupport/components/info/info.tmpl.html#L6

**Comment by mdale** [comment]
@ranyefet I see what you mean per the loop... still I think moving helpers to a namespace location that can be documented and used both in iframe plugins and templates is the way to go; or if exposing "formaters" have them share the same evaluation formater names; not create a new helpers object. So you could use "dateFormat" along side "trDateFormat" and we could have clean documentation about "formaters" in the player.katlura.com/docs/api page. 

```javasciprt
<span>Entry created at: <%=formater.trDateFormat(entry.createdAt);%></span>
```
where "formaters" ==  mw.KDPMapping.formatFunctions

ranyefet:
@mdale @itaykinnrot I've fixed the issues, now all formaters are register into ```mw.KDPMapping.formatFunctions```
Also, templates can use ```mw.formaters.myFormat``` or just ```formaters.myFormat```

Please review & merge

OrenMe:
Closing -> implemented #1428

PR #1374, opened 2014-12-29
2.25 jan 11 share

eitanavgil:
Ability to add a basic embed code to the share screen and to have the ability to override it

### DIFF
--- a/modules/KalturaSupport/components/share/Share.html
+++ b/modules/KalturaSupport/components/share/Share.html
@@ -1,35 +1,35 @@
 <!DOCTYPE HTML>
 <html>
 <head>
-	<title>Share Plugin</title>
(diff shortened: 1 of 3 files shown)


itaykinnrot:
@eitanavgil Please use kwidget-ps repo for this kind of commits + create a new share/embed as we're going to add this functionality soon - you dont want to re-test it...

Ronileco:
@eitanavgil , I cann't use the studio editor if I use the embed code in the boxes there.
It embeds the page instead of showing the string.

What we have now works well and gives the customer the ability they need.

@itaykinnrot , I believe this changes can be done on core, they are very basic and expand the core functionality and providing the embed option.

Can we talk about this tomorrow?

PR #1396, opened 2015-01-05
FEC-2587 #comment Sync sidebar and top/bottom bars display #time 4h

OrenMe:
Set height of sidebar to take in consideration top and bottom bars.
While sidebar is open - freeze component hover

### DIFF
--- a/modules/KalturaSupport/components/sideBarContainer.js
+++ b/modules/KalturaSupport/components/sideBarContainer.js
@@ -73,7 +73,14 @@
 					_this.getConfig("minDisplayWidth") <= _this.getPlayer().getWidth() &&
 					_this.getConfig("minDisplayHeight") <= _this.getPlayer().getHeight())){
 					_this.render = true;
-					_this.getComponent().css('height', _this.getPlayer().getVideoHolder().height());
+					//Sidebar height is player height without the top and bottom bars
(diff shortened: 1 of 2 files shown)


modules/KalturaSupport/components/sideBarContainer.js
```diff
@@ -73,7 +73,14 @@
...
 					_this.getConfig("minDisplayHeight") <= _this.getPlayer().getHeight())){
 					_this.render = true;
-					_this.getComponent().css('height', _this.getPlayer().getVideoHolder().height());
+					//Sidebar height is player height without the top and bottom bars
+					var height = _this.getPlayer().getHeight() -
+						_this.getPlayer().getControlBarContainer().height() -
+						_this.getPlayer().getTopBarContainer().height();
+					_this.getComponent().css('height', height);
+					//If topbar exist then add top value
+					_this.getComponent().css('top', _this.getPlayer().getTopBarContainer().height());
```
itaykinnrot: Can be null reference? if we dont have top bar?

OrenMe: Actually no - jquery will return an empty jquery elements wrapper array that has the height function - but I agree that it's better practice to include a check here.

PR #1399, opened 2015-01-05
FEC-2532 #comment fix the back and forth play/pause notifications issue ...

InbarItay:
...#time 1d
Fix hte endless looping of states.
fixes the problem with nexus devices.
the android level code should deal with the problems with the exoplayer for LG devices


itaykinnrot:
@InbarItay this is a big change - i dont want to add it to 2.25

PR #1401, opened 2015-01-05
Added custom embed code instead of custom embed src

Ronileco:
1. Please note that you have to url encode the custom embed code.


itaykinnrot:
Please use PS for this kind of commits - Thanks!

PR #1405, opened 2015-01-06
FEC-2647 --  Use the original ad duration

amirch1:
don't update video duration using the VPAID object remaining time + current time. Use the original ad duration instead.


itaykinnrot:
Merge to 2.26

PR #1406, opened 2015-01-06
FEC-2483 -- added volumeControl to  excluded list for disable plugins

amirch1:
added volumeControl to the excluded lest of disabled plugins (volume control is always enabled)


itaykinnrot:
Merged to 2.26

PR #1407, opened 2015-01-06
FEC-2662 -- kwidget-ps path templates

eitanavgil:


mdale:
we should make the pull against 2.25 or 2.26 ?

modules/KalturaSupport/UiConfResult.php
```diff
@@ -240,6 +243,7 @@ private function walkThroughPluginAttributes( $attrArray ,&$refrencePathInObject
 				$this->walkThroughPluginAttributes( $attrValue , $refrencePathInObject[$attrKey] );
 			}
 			else {
+			    if( in_array($attrKey, $this->propsToIgnoreWhenResolvingResource) ) continue;
```
itaykinnrot: @ranyefet can you explain why you're not using the {html5ps}?

eitanavgil: @itaykinnrot - when we DO use {html5ps} it doesn't work, with this small fix it does. I don't have Ran now, but I don't want to miss this merge. I think that specifically for templatePath it doesn't work (while other paths do)

Can you or someone from your team review this? (and make sure that this didn't break core templatPath) 
Client is an important one and was promised this for a while.

itaykinnrot:
@eitanavgil do you still need it?

PR #1408, opened 2015-01-07
FEC-2352 -- Default enabled plugins & support mp4 sources with kPlayer  in media inject 

mdale:
add more default enabled plugins & support mp4 sources with kPlayer  in media inject


mdale:
@MichalRadwantzor maybe we could take a quick look at this as well?

PR #1427, opened 2015-01-08
PS-2015 support omniture.plugin attribute & "content play" tracking fix ( 2.21.14 )

eitanavgil:


itaykinnrot:
Already in our latest

PR #1432, opened 2015-01-11
#SUP-3446 #comment watermark timeout display #time 2h

GuyLivni:


itaykinnrot:
@GuyLivni did you check playlist/changemedia?

itaykinnrot:
@mdale please approve this feature

modules/KalturaSupport/KalturaSupport.manifest.php
```diff
@@ -810,6 +810,14 @@
 				'initvalue' => 'http://www.kaltura.com/',
 				'type' => 'url'
 			),
+			'watermarkTimeout' => array(
```
mdale: "hideTimeout"
may be a better name since you already have watermark as the plugin key, you don't need to repeat the watermark  watermark prefix.

mdale:
approved but maybe update the attribute as mentioned.

GuyLivni:
Sure, updated the attribute name to hideTimeout

PR #1446, opened 2015-01-15
#FEC-2703 #comment enable wideinve when nativeCallout is set

MichalRadwantzor:


itaykinnrot:
@MichalRadwantzor lets add it to the next version

PR #1469, opened 2015-01-20
#SUP-3682 #comment fix logo external href opens in self window #time 2h

GuyLivni:


itaykinnrot:
Not going to happen - external integrator should use the plugin

GuyLivni:
I noticed that this href is coming externally from: $('.logo a').attr('href', current_domain.url);
Since they already load the href externally, and title as well in their custom plugin: custom-v1.js

I suggested that they add: $('.logo a').attr('target', "_blank");

PR #1475, opened 2015-01-21
FEC-2747 -- reset streamerType for each entry

amirch1:
reset streamerType for each entry according to the original streamerType Flashvar (if specified)

### DIFF
--- a/modules/KalturaSupport/resources/mw.KWidgetSupport.js
+++ b/modules/KalturaSupport/resources/mw.KWidgetSupport.js
@@ -114,7 +114,7 @@ mw.KWidgetSupport.prototype = {
 		
 		// Add hook for check player sources to use local kEntry ID source check:
 		embedPlayer.bindHelper( 'checkPlayerSourcesEvent', function( event, callback ) {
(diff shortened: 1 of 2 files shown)


modules/KalturaSupport/resources/mw.KWidgetSupport.js
```diff
@@ -113,6 +114,7 @@ mw.KWidgetSupport.prototype = {
 		
 		// Add hook for check player sources to use local kEntry ID source check:
 		embedPlayer.bindHelper( 'checkPlayerSourcesEvent', function( event, callback ) {
+			_this.originalStreamerType = embedPlayer.getKalturaConfig( null, 'streamerType' ) ? embedPlayer.getKalturaConfig( null, 'streamerType' ) : 'auto';
```
itaykinnrot: @amirch1 if we dont have streamer type why do we add auto?
the streamer type will tell the player to use kplayer and not html5 when possible

itaykinnrot:
Please add comment to QA what they need to check here

PR #1499, opened 2015-02-02
#SUP-3783 #comment fix show all captions and scroll #time 2h

GuyLivni:


itaykinnrot:
@GuyLivni did you check this fix on IE8/9/10 ipad and android tabled firefox,safari ?

GuyLivni:
works on all of the above, besides IE8, however, it did not work there even before the fix.

itaykinnrot:
@GuyLivni please check this link http://stackoverflow.com/questions/7707/ie8-overflowauto-with-max-height :-)

GuyLivni:
Hello @itaykinnrot , I checked the link, it seems like its a limitation of IE8 when you load too much data..
I tried the workarounds but they did not seem to take effect.

PR #1516, opened 2015-02-08
Kontiki plugin fixes

rkreich:
- Fixed Kontiki race condition between plugin load and kontiki agent response. Kontiki plugin will now wait for the Kontiki agent to respond before letting the player continue loading (asyncInit: true). Default timeout is set to 2 seconds.
- Explicitly set `document` when writing proxy swf.
- Added additional logging.

@itaykinnrot,
Please review and merge. Then we need to cherry pick it into 2.22.1


itaykinnrot:
@rkreich i've tagged v2.22.2 with this fix  - please verify

PR #1525, opened 2015-02-12
FEC-2918

eitanavgil:
Add uiconf id to Google Analytics label


eitanavgil:
@mdale  @itaykinnrot 
I added this as an additional argument to GoogleAnalytics label (last argument so we do not break existing data) 
Please tell me if you think this is OK with you or you need me to add a flag to add this or not.

PR #1532, opened 2015-02-16
FEC-2803 -- remove source unless its the only available flavor

amirch1:
remove source flavor from the sources array unless its the only available flavor


itaykinnrot:
@amirch1 we want to prefer mbr over web - we need to change the code

Issue #1563, opened 2015-02-24
VPAID Overlay not triggered on HTML5 Player (Mobile Only)


ocundale:
I have setup a simple demo here which works perfectly on desktop, with the overlay appearing after ~5 seconds:
http://coullwebsdk.herokuapp.com/minimal-test.php

However, the same demo does not work on mobile devices such as Android devices. The reason for this seems to be that the 'overlayUrl' (as specified in the kWidget embed code) is not hit at the specified time.
Interestingly, if you pause the video and then hit play again, the creative will then load and the overlay appears. Let me know if you require any further details!
Cheers :)

ocundale:
Also- It may save you some time to know that you can easily reproduce this using Chromes Mobile Device Emulation from the Developer Tools console.

amirch1:
Thanks a lot! This is a bug and we will fix it for the next version.

ocundale:
Brilliant, thanks for the update amirch!

jessp01:
Closing issue. Thanks again @ocundale

PR #1581, opened 2015-03-03
FEC-3115 -- run VPAID flow for overlay on mobile

amirch1:
run VPAID flow regularly when the ad is an overlay on mobile device


mdale:
@amirch1  does this address: https://github.com/kaltura/mwEmbed/issues/1563

PR #1596, opened 2015-03-09
add live support for mediaproxy + add indication that we're working with...

itaykinnrot:
...out platform


itaykinnrot:
We have... - look up

PR #1600, opened 2015-03-09
use $this->getEntryResult() check instead of request checks.

mdale:


mdale:
merging because I accidentally put the native doc update into this commit, and the change is pretty safe ( eclipse seems to be not displaying all commit data )

PR #1662, opened 2015-03-25
#SUP-4030 #comment doUpdateLayout when player ready #time 2h

GuyLivni:


modules/EmbedPlayer/resources/mw.EmbedPlayer.js
```diff
@@ -1321,6 +1319,9 @@
 			$(this).trigger('playerReady');
 			this.triggerWidgetLoaded();
 
+			// Update layout
+			this.doUpdateLayout();
+
```
amirch1: @GuyLivni @itaykinnrot  - updateLayout must come before we can trigger playerReady. We expect the DOM to be ready when the playerReady event is triggered. This change caused a regression: FEC-3389. Please advise.

PR #1668, opened 2015-03-26
fix Widvine on Andoroid + AutoPlay - Do not merge!!!

InbarItay:


modules/Widevine/resources/mw.Widevine.js
```diff
@@ -86,6 +86,7 @@
 
 					//if mobile device
 					if ( kWidget.isMobileDevice() ) {
+						return;
```
itaykinnrot: Need to add a check if native callout is active and not return

PR #1670, opened 2015-03-29
FEC-3125

amirch1:
pause video after change media on mobile devices


amirch1:
Testing guidelines:
1. Check changeMedia call by API on all devices, with and without ads (switch during ad playback).
2. Verify playlists continue to change media as expected
3. Verify source switch (using the source selector) works as expected

PR #1681, opened 2015-03-30
#FEC-3133 #comment removePoster after changeMedia #time 2h

GuyLivni:


modules/EmbedPlayer/resources/mw.EmbedPlayer.js
```diff
@@ -1714,11 +1714,12 @@
 					// Reset changeMediaStarted flag
 					_this.changeMediaStarted = false;
 
+					if (!_this.isAudioPlayer) {
```
itaykinnrot: We only want to remove the poster if we're going to play after the switch - please revert this change

PR #1711, opened 2015-04-13
Q and a plugin2

asafrob:
Q&A Plugin initial version

### DIFF
--- a/modules/QnA/resources/qna.js
+++ b/modules/QnA/resources/qna.js
@@ -68,7 +68,7 @@
 
 			this.bind('layoutBuildDone ended', function (event, screenName) {
 				// add the Q&A toggle button to be on the video
-				embedPlayer.getVideoHolder().append('<div class="qna-on-video-btn icon-qna-close"><div class="badge"></div></div>');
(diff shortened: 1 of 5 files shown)


modules/QnA/resources/qna.js
```diff
@@ -0,0 +1,200 @@
...
+
+		addBindings: function () {
+			var _this = this;
+			var embedPlayer = this.getPlayer();
+
+			this.bind('layoutBuildDone', function (event, screenName) {
+				// add the Q&A toggle button to be on the video
+				embedPlayer.getVideoHolder().append('<div class="qna-on-video-btn icon-qna-close"></div>');
+				_this.getQnaContainer();
+				var qnaObject =  $(window['parent'].document.getElementById(embedPlayer.id )).parent().find( ".qnaInterface" );
```
itaykinnrot: we must check that the player is not in unfriendly iframe - this code will break if we do

modules/QnA/resources/qna.js
```diff
@@ -0,0 +1,200 @@
...
+				var onVideoTogglePluginButton = $('.qna-on-video-btn');
+				// register to on click to change the icon of the toggle button
+				$(".qna-on-video-btn").on("click", function(){
```
itaykinnrot: use onVideoTogglePluginButton

PR #1718, opened 2015-04-15
SUP 4213 preferedFlavorBR

einatr:
assigned to Michael D. for review


mdale:
I don't think that firefox code is needed now that firefox support h.264 on all platforms.

PR #1725, opened 2015-04-20
FEC-3390

amirch1:
set seeking to false after YouTube seekTo (as YouTube doesn't dispatch a "seeked" event)


amirch1:
@OrenMe  - the regression is caused due to this change:
https://github.com/kaltura/mwEmbed/blob/master/modules/EmbedPlayer/resources/mw.EmbedPlayer.js#L2183
Since YouTube doesn't dispatch a "seeked" event we must turn "seeking" back to false after calling Youtube "seekTo" method in order to avoid entering the condition in the code above.

PR #1730, opened 2015-04-20
#SUP-4003 #comment captions selection logic

GuyLivni:


GuyLivni:
@mdale @itaykinnrot  please check that it looks ok. Added captions logic selection info.

PR #1731, opened 2015-04-20
service plugin to inject playlist at runtime

eitanavgil:


mdale:
Thanks, looks like a good addition to the playlist API. 

Where is the testfile and API documentation update? 

We can't add or merge API updates without documentation and test files documenting respective additional events. We presently don't have a good place for plugins notifications tracking. will see if there is something we can do about that .. but in the mean time a test page is good for now :)

amirch1:
merged

PR #1735, opened 2015-04-21
#SUP-4030 #comment update layout after hideComponents #time 1h

GuyLivni:


itaykinnrot:
@GuyLivni  open it to the next version

PR #1740, opened 2015-04-23
FEC-3421 Don't triger clipDone while changing media

mdale:


mdale:
was fixed another way ( right @amirch1 ? )

PR #1768, opened 2015-05-04
AdaptiveBitrate SourceSelector

einatr:
#FEC-2897,FEC-369 #comment added adaptive bitrate support for flavour
selector #time 3d


itaykinnrot:
@einatr how did you solve the back to auto?

itaykinnrot:
@einatr need to check with silverlight and dash @OrenMe

PR #1801, opened 2015-05-13
2.31 may 31

itaykinnrot:


mdale:
merging to master as this is on production, and there is examples from this release that I need to share with the player.kaltura.com URLs

PR #1817, opened 2015-05-21
SUP-4277

amirch1:
remove corner-radius css property for windows phone as it breaks. use click event instead of touchStart event as it doesn't work on windows phone.


modules/KalturaSupport/components/largePlayBtn.js
```diff
@@ -101,6 +101,9 @@
 							.on(eventName, function(e) {
 								_this.clickButton(e);
 							} );
+				if (!mw.isWindowsPhone()){
+					this.$el.css("border-radius", "0.1em");
```
itaykinnrot: We should add it as css and not hardcoded - think of someone doing their own skin....

PR #1820, opened 2015-05-25
debugInfo

kalturaguy:
Ability to view player debug information (in future we can put there which HLS segment is downloaded, which one is player, current player buffer size, etc..)

### DIFF
--- a/modules/DebugInfo/resources/DebugInfo.js
+++ b/modules/DebugInfo/resources/DebugInfo.js
@@ -1,0 +1,167 @@
+( function( mw, $ ) {"use strict";
+
+mw.PluginManager.add( 'debugInfo', mw.KBaseComponent.extend({
+
+	defaultConfig: {
+		templatePath: '../debugInfo/resources/DebugInfo.tmpl.html',
+        onPage: true,
+        cssFileName: 'modules/debugInfo/resources/DebugInfo.css'
+	},
+    getBaseConfig: function() {
+        var parentConfig = this._super();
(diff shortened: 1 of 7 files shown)


modules/DebugInfo/resources/DebugInfo.js
```diff
@@ -0,0 +1,167 @@
...
+                    internal[name]=_this.$scope[name];
+
+                    function updateHtml() {
+                        var newContent = originalText.replace(match, internal[name]);
+                        el.innerHTML = newContent;
+                    }
+
+                    updateHtml();
+
+                    Object.defineProperty($newscope, name, {
```
itaykinnrot: this will crash on IE8

modules/DebugInfo/resources/DebugInfo.js
```diff
@@ -0,0 +1,167 @@
...
+                        },
+                        set: function(newValue) {
+                            try {
+                                if (internal[name] !== newValue) {
+                                    internal[name] = newValue;
+                                    updateHtml();
+                                }
+                            }
+                            catch(e) {
+                                alert(e);
```
itaykinnrot: alert?

kalturaguy: Fixed

modules/DebugInfo/resources/DebugInfo.js
kalturaguy: Removed support for IE8

PR #1844, opened 2015-06-01
ads and documents some simple properties for ad tag urls:

mdale:
* flashVersion
* playerVersion
* player width, height & parent page position.


mdale:
@itaykinnrot & team please review, is low risk addition, but needed per the example page referenced: 
http://kgit.html5video.org/pulls/1844/modules/AdSupport/tests/AdTagUrlSubstitutions.html

PR #1851, opened 2015-06-04
Multi audio test

itaykinnrot:


mdale:
Maybe we should issue a seek to the current time when the users switches audio streams to clear the buffer and support fast audio switch while not yet implemented natively ?

amirch1:
opening a new pull request to 2.33

PR #1858, opened 2015-06-09
KMS-7333 #comment remove the enable gui from flash - this should be h…

itaykinnrot:
…andled like the rest of our players #time 1h

### DIFF
--- a/modules/EmbedPlayer/resources/mw.EmbedPlayerKplayer.js
+++ b/modules/EmbedPlayer/resources/mw.EmbedPlayerKplayer.js
@@ -146,7 +146,6 @@
 						'switchingChangeStarted': 'onSwitchingChangeStarted',
 						'switchingChangeComplete': 'onSwitchingChangeComplete',
 						'flavorsListChanged': 'onFlavorsListChanged',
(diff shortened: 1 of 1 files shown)


modules/EmbedPlayer/resources/mw.EmbedPlayerKplayer.js
```diff
@@ -146,7 +146,6 @@
 						'switchingChangeStarted': 'onSwitchingChangeStarted',
 						'switchingChangeComplete': 'onSwitchingChangeComplete',
 						'flavorsListChanged': 'onFlavorsListChanged',
-						'enableGui': 'onEnableGui',
 						'liveStreamOffline': 'onLiveEntryOffline',
```
amirch1: @itaykinnrot - as suspected, this breaks UI disabling during ads when using KPlayer. We need to look into it and find why the JS disable doesn't kick-in before we can merge this.

amirch1: Seems like another issue - merging and checking the issue.

PR #1871, opened 2015-06-12
FEC-3674 -- Check for metadataProfileId flashvar, include if set

mdale:


mdale:
please review, we need this to be part of an upcoming release.

PR #1888, opened 2015-06-16
FEC-3680 #comment Remove PHP definition files and loading

OrenMe:


itaykinnrot:
@OrenMe  please talk with @guy1976 & @atarsh @mrkfdr and instruct them to move to JSON for q/a & quiz before merging this one.

mdale:
@OrenMe something like this you want to merge right away because new plugins are being built all the time and will all need to be moved over ( all the time ) unless we have a hard switch against master for changes like this.

PR #1894, opened 2015-06-17
SUP-5120 -- correct currentTime before seek

amirch1:
get the correct currentTime before seek


modules/EmbedPlayer/resources/mw.EmbedPlayer.js
```diff
@@ -3082,7 +3082,7 @@
 			if (!this.isStopped()) {
 				this.isFlavorSwitching = true;
 				// Get the exact play time from the video element ( instead of parent embed Player )
-				var oldMediaTime = this.getPlayerElement().currentTime;
+				var oldMediaTime = this.currentTime;
```
itaykinnrot: i would change the remark or check the source of it. @amirch1

PR #1910, opened 2015-06-22
#SUP-5184 #comment set displayCaptions false when no sources #time 1h

GuyLivni:


itaykinnrot:
@GuyLivni  please reopen for 2.34 and assign @einatr  for review

PR #1914, opened 2015-06-23
FEC-3683

amirch1:
add and remove rotate class only when changing source / stream and not when enabling / disabling the button. Also fix sourceSwitchingEnd event triggering logic and allow source switching when the player is not playing (for example when the player is loaded and ready).

### DIFF
--- a/modules/EmbedPlayer/resources/mw.EmbedPlayer.js
+++ b/modules/EmbedPlayer/resources/mw.EmbedPlayer.js
@@ -3076,28 +3076,26 @@
 				{ currentBitrate: currentBR }
 			]);
 			this.mediaElement.setSource(source);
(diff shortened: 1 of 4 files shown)


amirch1:
@mdale - renamed "rotate" class name to "inProgress" - is this OK?

mdale:
Do we have any other camel case states? We should align naming conventions with existing CSS states: 
http://knowledge.kaltura.com/kaltura-player-toolkit-theme-skin-and-plugins-guide#Player States CSS <http://knowledge.kaltura.com/kaltura-player-toolkit-theme-skin-and-plugins-guide#Player%20States%20CSS>

amirch1:
@mdale  - how does "in-progress-state" sounds?

mdale:
Perfecto ;)

PR #1918, opened 2015-06-23
support anonymous analytics features:

eitanavgil:
hideKs: hides the KS from stats beacons even if embed-code received one
hideUserId: hides the userId from stats beacons even if embed-code received one


modules/KalturaSupport/tests/KalturaAnalytics.qunit.html
```diff
@@ -97,6 +97,8 @@
 			'originFeature': '0',
 			'applicationName': 'customAN',
 			'userId': 'customUid'
+//			'hideUserId': true,
+//			'hideKs' : false
```
mdale: you should add these to the page with their defaults ( i.e hideUserId false  and hideKs false ) 

You should also document these new properties in the kalturaSupport.manifest.json file so that they can have player studio and feature config representation.

PR #1919, opened 2015-06-23
#SUP-5206 #comment init CC plugin with visibility false #time 1h

GuyLivni:


itaykinnrot:
@GuyLivni  please reopen for 2.34 and assign to @einatr  for review

PR #1929, opened 2015-06-28
Hls debug events

einatr:
hls debug info support added
HLS test page updated with new values


modules/DebugInfo/resources/DebugInfo.js
```diff
@@ -0,0 +1,148 @@
...
+        });
+
+       // _this.setVisible(true);
+        $(document).keydown(function(e){
+
+            if ( e.altKey && e.ctrlKey && e.keyCode===68) {
+               _this.setVisible(!_this.isVisible);
+
+            }
+        })
```
itaykinnrot: missing ;

PR #1938, opened 2015-07-01
Caption as Service - iPhone support

ranyefet:
Hi,

This pull request fixes problem when ```forceLoadLanguage``` is enabled but not working on iPhone.
I've tested the captions on iPhone with and without the fix and everything is still working so it shouldn't cause issues.

@eitanavgil @itaykinnrot 
Please review & merge.


eitanavgil:
+ @mdale FYI

eitanavgil:
@itaykinnrot are you following this?

eitanavgil:
@shomronk  - Is this under your radar?

PR #1943, opened 2015-07-03
FEC-2900 -- use "ended" not onEndDone

mdale:


modules/PlayersJsReceiver/resources/mw.PlayersJsReceiver.js
```diff
@@ -77,7 +77,8 @@
 					'duration': player.duration
 				});
 			});
-			this.bind('onEndedDone', function(){
+			// use "ended" not onEndDone, because when loop is active onEndDone is not triggered.
+			this.bind('ended', function(){
 				receiver.emit('ended');
```
amirch1: @mdale - the "ended" event triggers before postrolls. Please consider using the "postEnded" event instead as it always triggers at the end of either the video or the postroll if exists. It will also work if "loop" is set to true and with chromeless player as well (tested).

mdale: Sounds good.

PR #1960, opened 2015-07-06
(Template) Formatters refactor

ranyefet:
Hi guys,

I had a race condition issue with adding my custom formatter.
Until now we used to store the format functions on the KDPMapping prototype object and expose it as ```mw.formaters``` object.

In my plugin I tried to register my formatter early as possible because the formatter is called early on.
I wasn't able to register it early because KDPMapping only exposed it later on.

So I moved the storage for the formatters into the same place where we expose the ```mw.util.registerTemplateHelper``` method which makes more sense.

I also noticed that there are calls to mw.formaters and KDPMapping.prototype in some plugins which is __bad__. 
So now ```mw.formaters``` is a function with the following API:
```JS
formaters: function() {
	return {
		register(name, callback),
		get(name),
		getAll(),
		exists(name),
	}
}
```

I've refactor all the places that used to access the formatters object directly while still maintaining the old way to register formatters using ```mw.util.registerTemplateHelper``` with deprecation message.

@itaykinnrot @eitanavgil @mdale 

Please review & merge


OrenMe:
@ranyefet I think we should make the template formatters a part of the mw.util.tmpl namespace.
e.g. mw.util.tmpl.formaters as they are an integral part of the template engine anyway.
What do you think? Are we using them else where?

eitanavgil:
@OrenMe  - we are. I have a few other places (in code) that are not in the templates.

eitanavgil:
@ranyefet @OrenMe can you please communicate and understand if there is a gap or not and resolve this?

OrenMe:
@ranyefet , @eitanavgil  if this is a general utility belt then please place this in **mediawiki.util.js** - it would be more intuitive to understand this is a general tool and not template only related.

ranyefet:
@OrenMe basically the formatters are just collection of functions. 
We use them within templates and also in evaluation, i.e:
```
player.evaluate('{mediaProxy.entry.createdAt|dateFormat}');
```
That's why I used a top level property on ```mw``` object.

Do you want to move it to somewhere else?

PR #1977, opened 2015-07-10
FEC-1246. Add browser/OS/features detection

vladm3:
FEC-1246. Add browser/OS/features detection


eitanavgil:
@mdale, @itaykinnrot - this is for https://kaltura.atlassian.net/browse/FEC-1246 

This will help us a lot with some support tickets that has some specific issues per device (like if we have a fullscreen issue on IE 9, we would be able to hide the button with CSS with this feature) 

I don't see an actual pull folder on kgit with 1977 probably because we forked it, this is why we don't have an quick test page - but this should be reflected on any player out there with this code. 

Please review and merge

mdale:
@eitanavgil can you set as same account pull request ?  this will register it in jira, give us the staging pages etc.

PR #2020, opened 2015-07-24
add carousel automatic scrolling when media changes

vladm3:


eitanavgil:
This is scrolling the Carusel to show the played entry automatically (like if there is an auto-play-next-item, the playlist will show the next item and not play 'invisible items' 

@itaykinnrot @mdale  @shomronk  - when can you review this? 
@amirch1 FYI

PR #2028, opened 2015-07-28
Etag enhancements

matsuu:
This patch fixes CDN caching problem if CDN support 'Vary: Accept-Encoding'.

Apache deflates response body if clients set Accept-Encoding and Apache doesn't rewrite ETag header because backend generates that. So deflated response and non-deflated response has same ETag. it happened that CDN servers always return deflated response even if client has no Accept-Encoding.
Weakness indicator is used in this situation. See also RFC7232 section 2.3

### DIFF
--- a/modules/KalturaSupport/kalturaIframe.php
+++ b/modules/KalturaSupport/kalturaIframe.php
@@ -18,7 +18,7 @@ $kIframe = new kalturaIframeClass();
 if(!ob_start("ob_gzhandler")) ob_start();
 
 // Support Etag and 304
-if( $wgEnableScriptDebug == false && @trim($_SERVER['HTTP_IF_NONE_MATCH']) == $kIframe->getIframeOutputHash() ){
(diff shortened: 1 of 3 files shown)


OrenMe:
@matsuu can you please add some additional information regarding the weakness indicator. Thanks.

jessp01:
@matsuu - can you please reply? thanks.

matsuu:
https://tools.ietf.org/html/rfc7232#section-2.3
https://tools.ietf.org/html/rfc7232#section-2.3.3

jessp01:
@OrenMe - please review.

OrenMe:
@matsuu thanks for the added info.
I'm not clear where the issue is at? the etag code changes won't change anything on our side as I understand as we do simple string comparing which will be true with the addition of the weak flag (/W) or without.
If I understand correctly you are referring to CDN caching issue.
Are you using our SaaS or an onPrem version?

OrenMe:
Close due to inactivity.
Please reopen if relevant.
@matsuu would appreciate your comment.

PR #2030, opened 2015-07-29
FEC-3814

amirch1:
open social network in a new tab instead in a window on Microsoft Edge only


amirch1:
verified this issue doesn't exists on MS Edge final release

PR #2035, opened 2015-07-30
#SUP-5481 #comment enable control bar for audio player native #time 2h

GuyLivni:
- Enabled control bar for native audio player.
- Check if isAudioPlayer player when you load and update player data.
- Fix audio player test page height.

### DIFF
--- a/modules/KalturaSupport/components/controlBarContainer.js
+++ b/modules/KalturaSupport/components/controlBarContainer.js
@@ -9,8 +9,8 @@
 		keepOnScreen: false,
 
 		setup: function(){
-			// Exit if we're using native controls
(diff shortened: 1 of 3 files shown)


itaykinnrot:
@GuyLivni  what do you want to do here?

GuyLivni:
I want to show the control bar for native device.
For example on iPhone what you see is black square with small strange audio symbol.
Its bad user experience, and seeing the control bar, which is functional is much better.
But as @mdale  mentioned audio tag is still required to not open the audio file in full screen.
So I guess this is part of FEC-2740, unless the audio tag/player there will do the same, then this can be disregard and closed.

itaykinnrot:
I'm going to close it

PR #2081, opened 2015-08-12
PLAT-3588 timeAlignedRenditions added

einatr:


modules/EmbedPlayer/resources/mw.EmbedPlayerKplayer.js
```diff
@@ -802,7 +806,11 @@
 				+ ksString + "/uiConfId/" + this.kuiconfid + this.getPlaymanifestArg("referrerSig", "referrerSig")
 				+ this.getPlaymanifestArg("tags", "flavorTags") + "/a/a." + fileExt + "?referrer=" + this.b64Referrer;
 
-			if (srcUrl.indexOf("&seekFrom=") !== -1) {
+            // add timeAlignedRenditions indicator in the case of HLS
```
einatr: Not sure this addition is needed. Seems like we never build a playManifest for a live stream...

PR #2085, opened 2015-08-12
change MS IE detection method

vladm3:


rkreich:
Quick background: There's an issue with the player when embedded in a windows application that uses the installed IE as a "white-labled" browser.

The user agent is taken from the IE installed, but it doesn’t take into account the compatibility mode, which is required for another part of the application.

For example, if the user has IE 10, the UA in windows application is:
`Mozilla/5.0 (compatible; MSIE 10.0; Windows NT 6.1; Trident/6.0)`
(which is the same as in real IE10 in the default state and Edge document mode), and when using compatibility mode, a real browser adjusts UA string to tell web-pages about this, so it becomes:
`Mozilla/4.0 (compatible; MSIE 8.0; Trident/4.0)`. 
In the windows application it isn’t adjusted and it’s left the same as in normal mode, so the player can’t get the real document mode from the UA. 

@eitanavgil

eitanavgil:
@mdale @itaykinnrot  Can we please push on this? TR just escalated this today.

eitanavgil:
Since this pull did not create a kgit version I created a new pull instead: 

https://github.com/kaltura/mwEmbed/pull/2102

Closing this pull request - please refer to the new one instead

PR #2089, opened 2015-08-13
Checking for XML Element in Vast ad XML instead of generic node

abaylis:
If the first XML node in a Vast feed is a comment, the getAdFormat function fails to idenitfy the format as 'vast', even when the feed is valid.  If ad type is chosen by element name, it should instead only check names if the node is an element.  

Example Feed:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!-- comment here -->
<VAST version="2.0">
	<Ad id="1">
	...
	</Ad>
</VAST>
```

### DIFF
--- a/modules/AdSupport/resources/mw.AdLoader.js
+++ b/modules/AdSupport/resources/mw.AdLoader.js
@@ -97,20 +97,42 @@ mw.AdLoader = {
 	 * 		The type of string
 	 */
 	getAdFormat: function( xmlObject ){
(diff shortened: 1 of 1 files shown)


amirch1:
@abaylis - sorry, this was missed from the last version. Can you please reopen on 2.36-Oct-20 and I will merge it? Thanks!

abaylis:
re-opened as https://github.com/kaltura/mwEmbed/pull/2179

PR #2091, opened 2015-08-16
Chromecast receiver

amirch1:


modules/Chromecast/resources/chromecast.js
```diff
@@ -447,20 +479,19 @@
...
 			// find the best quality MP4 source
 			var sources = this.embedPlayer.mediaElement.sources;
-			var requiredMimetype = "video/mp4";
 			var videoSize = 0;
 			var newSource = null;
 			var i = 0;
 			for ( i=0 ; i < sources.length; i++){
 				var source = sources[i];
-				if (source.mimeType === requiredMimetype && parseInt(source.sizebytes) > videoSize){
+				if ((source.mimeType === 'video/mp4' || source.mimeType === 'application/dash+xml') && parseInt(source.sizebytes) > videoSize){
```
itaykinnrot: @amirch1 we should check HLS too for live content (can be phase 2)

PR #2097, opened 2015-08-17
#SUP-5618 #comment hide playhead out of player #time 2h

GuyLivni:


amirch1:
@GuyLivni  - I see some strange behaviour in iPAD. Lets take a look together before merging...

PR #2100, opened 2015-08-19
#FEC-4038 added and documented enhanced buffer events

mdale:
* Added and documented enhanced buffer events
* Added with working example against GA plugin.


itaykinnrot:
Merged to 2.35

PR #2106, opened 2015-08-23
#SUP-5635 #comment fix font size for elements #time 1h

GuyLivni:


itaykinnrot:
What about the big play button, @amirch1 please consult here.

amirch1:
@itaykinnrot we don't want to affect the large play button with this settings so this is ok.

PR #2118, opened 2015-08-27
2.33.3 ie document mode

rkreich:


eitanavgil:
@itaykinnrot  please see when and how we can review this. It fixes an IE8 issue with TR.

eitanavgil:
created a parallel pull request against 2.35. Please use https://github.com/kaltura/mwEmbed/pull/2135 instead

rkreich:
@eitanavgil #2135 doesn't have the right commits in it 
https://github.com/kaltura/mwEmbed/pull/2135/files

PR #2134, opened 2015-09-04
core part of TR-330

vladm3:
Add empty playlist handling


eitanavgil:
@vladm3  I think the fix doesn't work. 
I applied it to a more advanced branch (on 2.35.sep-20) and also added the attribute to the kbaseMediaList (if there is a reference from kbaseMediaList to the new attribute it must be implemented it as well and not on classes that extends it) and I see that this is failing to hide the error alert. 
The new branch name is TR-331. 
New Branch pull is https://github.com/kaltura/mwEmbed/pull/2135. Please work on that branch. The effort is not to show the error and not only on the 1st load of the playlist.

PR #2135, opened 2015-09-05
Tr 331

eitanavgil:
@itaykinnrot  please review and merge

### DIFF
--- a/modules/KalturaSupport/components/playlistAPI.js
+++ b/modules/KalturaSupport/components/playlistAPI.js
@@ -34,7 +34,9 @@
 			'includeHeader': true,
 			'renderWhenEmpty': false,
 			'paging': false,
-			'pageSize': 25
+			'pageSize': 25,
+			'showEmptyPlaylistError': true
+
 		},
 
 
@@ -649,7 +651,7 @@
 				} );
 			}
 
-			if (items.length === 0){
+			if (items.length === 0 && (this.firstLoad || this.getConfig('showEmptyPlaylistError'))) {
 				//If no items then show error message
 				this.showEmptyPlaylistError();
 				this.configMediaListFeatures();
--- a/modules/KalturaSupport/components/playlistAPI.js
+++ b/modules/KalturaSupport/components/playlistAPI.js
@@ -34,8 +34,8 @@
 			'includeHeader': true,
 			'renderWhenEmpty': false,
 			'paging': false,
-			'pageSize': 25
-
+			'pageSize': 25,
+			'showEmptyPlaylistError': true
 		},
 
 
@@ -650,7 +650,7 @@
 				} );
 			}
 
-			if (items.length === 0 && (this.firstLoad || this.getConfig('showEmptyPlaylistError'))) {
+			if (items.length === 0 && this.getConfig('showEmptyPlaylistError')) {
 				//If no items then show error message
 				this.showEmptyPlaylistError();
 				this.configMediaListFeatures();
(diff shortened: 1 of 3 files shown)


modules/KalturaSupport/components/playlistAPI.js
```diff
@@ -34,7 +34,8 @@
 			'includeHeader': true,
 			'renderWhenEmpty': false,
 			'paging': false,
-			'pageSize': 25
+			'pageSize': 25,
+			'showEmptyPlaylistError': true
 		},
```
amirch1: @eitanavgil  - no need to define it here if also defined in mw.KBaseMediaList.js as playlistAPI inherits it.

modules/KalturaSupport/resources/mw.KBaseMediaList.js
```diff
@@ -268,7 +268,7 @@
 		//Media Item
 		renderMediaList: function(callback){
 			//Only render if medialist item are present
-			if (this.getTemplateData().length > 0) {
+			if (this.getTemplateData().length > (0 - !this.getConfig('showEmptyPlaylistError'))) {
 				//Generate new list template data
```
amirch1: @eitanavgil  - I don't understand what you are trying to do here: the expression in brackets will always be 0 or -1.

eitanavgil: We still want the playlist template to be rendered so we can show our UI on the playlist area. For example if you add this to your playlist template we will show a custom UI when there is no loaded playlist: 
<% if (!mediaList || !mediaList.length) { %>
    <div class="ri-playlist-empty-error-wrap"> EMPTY RESULTS</div>
<% } %>

vladm3: this code is needed because we need to render the template even if the length of media items is zero, because the empty playlist error is handled inside the template

eitanavgil:
@amirch1 if this is not accepted please tell us we will think of a new approach.

amirch1:
@eitanavgil - I have no problem with this approach. But since 2.35 is way pass its code freeze date and I feel that this change must be verified to make sure it doesn't cause any other issues, It would be best to add it to 2.36. If this is too late for you - please let me know and I will consult with @itaykinnrot if we can push it to 2.35 on Sunday morning.

modules/KalturaSupport/resources/mw.KBaseMediaList.js
amirch1: @vladm3 - while this line does what is should, it is very unclear. Please try to avoid using boolean values inside arithmetic calculations, even if it means adding another line of code for checking the boolean condition. This will ensure readability in the feature. Thanks.

amirch1:
@eitanavgil  - can you please reopen this pull request on 2.36-Oct-20 and I will merge it? Note that I've asked @vladm3 to separate the arithmetic calculation and the boolean value check to 2 separate lines in order to make sure it is clear and readable in the future.

modules/KalturaSupport/resources/mw.KBaseMediaList.js
```diff
@@ -268,7 +268,7 @@
 		//Media Item
 		renderMediaList: function(callback){
 			//Only render if medialist item are present
-			if (this.getTemplateData().length > 0) {
+			if ( this.getTemplateData().length > 0 || ( this.getTemplateData().length === 0 && !this.getConfig('showEmptyPlaylistError') ) ) {
 				//Generate new list template data
```
amirch1: @eitanavgil , @vladm3  - I've changed the condition for better readability - please verify the logic works as it should after this change.

vladm3: @amirch1 it works correctly

PR #2143, opened 2015-09-07
fix email sharing (TR-194)

vladm3:
fix email sharing for cases when there’s no email client


eitanavgil:
@itaykinnrot  @mdale  (@amirch1) 
one of our clients found out that if there is no default email client on the PC there is an error - this pull fixes that. Please review and merge.

amirch1:
@itaykinnrot  - no risk here. Do you want to add this to 2.35 or 2.36 (as 2.35 already in code freeze)?

PR #2151, opened 2015-09-10
fix entry name title attribute not updated when component width less than availableWidth (TR-192)

vladm3:


eitanavgil:
@itaykinnrot @amirch1 this fixes https://kaltura.atlassian.net/browse/TR-192 please review

amirch1:
@eitanavgil - as far as I recall we wanted the tooltip only if the text is truncated. With this fix, there will always be a tooltip. I think a better fix will be to remove the tooltip on the playerReady event and keep the code which adds it when the text is truncated.

eitanavgil:
@vladm3  can you please make the changes so it behaves as Amir described?

vladm3:
@eitanavgil As I said before it's not easy to flawlessly develop the desired behaviour. But it's relatively easy to make the desired behaviour but with static title attribute.

amirch1:
@eitanavgil , @vladm3 - what's the status here? Are you going to change the code or you want to keep it as it is?

eitanavgil:
Amir, I added a URL and instructions on the Jira ticket. Please see there

eitanavgil:
https://amers1.app.kaltura.cp.icp2.mpp.reutest.com/kwf/widget/cmw/theme/nvs/instanceid/1_7ii9kgnr/mid/1438346 
you can see it here if you click on an entry with long name and then play with player-mode button

kaltura-hooks:
Hi @vladm3,

PR #2154, opened 2015-09-14
fix playlist item bad width (TR-331)

vladm3:
the description below is related to TR-331:

If the empty playlist error happened, mw.KBaseMediaList.js sets config MinClips to zero (because playlist is empty).
Then, when switching to "Player mode”, redrawPlaylist function is called, and here: https://github.com/kaltura/mwEmbed/blob/5d9e19ee5fff0ad23b595b6a6a508676cdf59856/modules/KalturaSupport/components/playlistAPI.js#L312 when updating mediaItemWidth division-by-zero takes place, so mediaItemWidth now equals Infinity, so all other math operations are screwed up and all this results in the TR-331.


eitanavgil:
@itaykinnrot this is a critical bug. Please review and merge

PR #2155, opened 2015-09-15
Passing through ajaxOptions for Wrapped Vast Feeds

abaylis:
ajaxOptions (i.e. enableCORS) are only used for the first vast feed in a request chain.  If the vast response contains a Wrapper ad stating to download another vast URL, the enableCORS option is not utilized in that secondary request.  Without these options, the initial request might have enableCORS set to true and be able to set permanent cookies, but secondary requests are unable to store permanent cookies.

There is only one global enableCORS config setting, it should be used on all requests.

Fix by passing through the ajaxOptions used in the initial request into the mw.VastAdParser.parse function so that they can be used in secondary vast URL requests initiated from within.

### DIFF
--- a/modules/AdSupport/resources/mw.AdLoader.js
+++ b/modules/AdSupport/resources/mw.AdLoader.js
@@ -51,7 +51,7 @@ mw.AdLoader = {
 			url: adUrl,
 			ajaxOptions: ajaxOptions,
 			success: function( resultXML ) {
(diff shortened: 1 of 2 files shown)


amirch1:
Thanks @abaylis - nice fix!
Any chance you can re-open this pull request on branch 2.36-Oct-20?

abaylis:
Sure thing, will close here and reopen there.

PR #2161, opened 2015-09-16
#SUP-5575 #comment fix utmp unfriendly iframe path #time 2h

GuyLivni:


modules/GoogleAnalytics/resources/mw.GoogleAnalytics.js
```diff
@@ -105,6 +105,9 @@
 			if( this.getConfig( 'anonymizeIp' ) ){
 				window._gaq.push(['_gat._anonymizeIp']);
 			}
+			if ( !mw.getConfig('EmbedPlayer.IsFriendlyIframe' ) ){
+				window._gaq.push(['_set', 'page', document.referrer.replace(/^[^:]+:\/\/[^/]+/, '').replace(/#.*/, '').replace(/\?.*/, '')]);
```
itaykinnrot: @GuyLivni  please add some checks before replacing so many strings

GuyLivni:
@itaykinnrot  I added check that document.referrer is defined.
If any additional checks are required please tell me which, the replaces are to create a path for utmp, first replace is to remove the protocol, second replace is to remove everything after "#", third replace is to remove everything after "?".

PR #2163, opened 2015-09-16
#SUP-5910 #comment fix HidePosterOnStart/ShowPosterOnStop collision #time 2h

GuyLivni:


OrenMe:
@GuyLivni - ended is not a standard property on all playback elements.
Please consider using the player abstraction for the playback state.

GuyLivni:
@OrenMe sure, I can use this.currentState=="end" instead, works as well. please review

OrenMe:
@GuyLivni you have duplicate code, please refactor it so base class returns true/false and check in derived class the state and exit accordingly.

You call base class from derived class at mw.EmbedPlayerNative.js@525, and then you run the same code in base class, so just return the state there and act accordingly.

PR #2167, opened 2015-09-17
Allowing Impressions from multiple vast wrappers

abaylis:
If a vast request chain results in more than one wrapper ad, e.g.

wrapper 1 -> wrapper 2 -> ad

then tracking impressions found on all wrappers mut be sent when the main ad's tracking URLs are called.  

But currently only tracking urls found on the final wrapper (e.g. wrapper 2) are being called.  This is because the wrapperData variable is cleared each time the mw.VastAdParser.parse function is called, but the function has to be called for each vast response.

Fix - Instead of storing only the last wrapperData, store all previous wrapperData objects in an array, and when setting up tracking URL-calling functionality, loop over all stored wrapperData objects and call each object's respective tracking URLs, not just those in the final wrapper.

### DIFF
--- a/modules/AdSupport/resources/mw.AdLoader.js
+++ b/modules/AdSupport/resources/mw.AdLoader.js
@@ -51,7 +51,7 @@ mw.AdLoader = {
 			url: adUrl,
 			ajaxOptions: ajaxOptions,
 			success: function( resultXML ) {
-				_this.handleResult( resultXML, callback );
+				_this.handleResult( resultXML, callback, ajaxOptions );
 			},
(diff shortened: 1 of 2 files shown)


abaylis:
Note - I believe the large number of line changes are just due to differnce in line endings,

mdale:
@abaylis would it be possible to clean the diff to avoid the line ending change that makes the changes to mw.VastAdParser.js a little hard to review.

abaylis:
Corrected and updated as https://github.com/kaltura/mwEmbed/pull/2180

PR #2178, opened 2015-09-21
Changed deep linking

nissopa:


itaykinnrot:
@mdale this is only for testing the solution

PR #2185, opened 2015-09-22
fix Stack Overflow in IE8 when changing media (TR-272)

vladm3:
May happen because IE8 has call stack limited to 12 frames (function calls)

### DIFF
--- a/modules/EmbedPlayer/resources/mw.EmbedPlayer.js
+++ b/modules/EmbedPlayer/resources/mw.EmbedPlayer.js
@@ -1677,7 +1677,9 @@
 			}
 
 			// Add a loader to the embed player:
-			this.pauseLoading();
+			this.isPauseLoading = true;
+			this.pause();
+			this.addPlayerSpinner();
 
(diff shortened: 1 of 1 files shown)


modules/EmbedPlayer/resources/mw.EmbedPlayer.js
```diff
@@ -1677,7 +1677,9 @@
 			}
 
 			// Add a loader to the embed player:
-			this.pauseLoading();
+			this.isPauseLoading = true;
+			this.pause();
+			this.addPlayerSpinner();
```
amirch1: While this works in a specific use-case, it means we are very close to the stack being full. If any other plugins that register to events triggered by the pause() and addPlayerSpinner() methods will be added to the player, I guess the stack overflow alert will return. Anyhow, I will close this pull request and reopen on 2.38 but we might have to handle it in a different way if the problem return.

amirch1:
Merged: https://github.com/kaltura/mwEmbed/pull/2347

PR #2186, opened 2015-09-23
fix playlist paging for externally loaded playlists with possible custom params (TR-408)

vladm3:


modules/KalturaSupport/components/playlistAPI.js
```diff
@@ -190,6 +192,14 @@
 						'pager:pageSize': _this.getConfig('pageSize'),
 						'id': _this.playlistSet[_this.currentPlaylistIndex].id
 					};
+
+					// extend with paging props
+					$.extend(playlistRequest, {
```
amirch1: @vladm3 - this code is redundant if not using saved params (when creating a new request object).

modules/KalturaSupport/components/playlistAPI.js
```diff
@@ -182,7 +182,9 @@
 					_this.pagingInProgress = true;
 					_this.page++; // move to next page
 					mw.log("Playlist:: scrollEnd event. paging: true: trying to load next playlist page: page="+_this.page+", pageSize"+_this.getConfig('pageSize'));
-					var playlistRequest = {
+
```
amirch1: @vladm3 - for readability and prevent redundant $.extend call, please separate the 2 requests creation:

if ( _this.playlistSet[_this.currentPlaylistIndex].playlistParams){
// use saved params and use $.extend
}else{
// create the original request
}

modules/KalturaSupport/components/playlistAPI.js
vladm3: @amirch1 right, it is actually redundant when not using saved params

modules/KalturaSupport/components/playlistAPI.js
vladm3: @amirch1 It's possible to resolve redundant call if remove L190-L192, but, I'll update file to use branching, like you've mentioned

vladm3:
@amirch1 - updated PR

PR #2187, opened 2015-09-24
#FEC-3938 #comment supporting npvr & epg #time 1d

ElizaSapir:


modules/Tvpapi/resources/getLicensedLinks.js
```diff
@@ -75,23 +75,29 @@
...
 
 				var successHandler = function ( res ) {
-					if( res.Status && res.Status.Code !== 0 ) {
-						_this.getPlayer().triggerHelper('tvpapiNoSubscription', [res]);
-					} else {
+					var url = getResponseLink(res);
+					if( url ) {
 						_this.getPlayer().triggerHelper('tvpapiSubscription', [res]);
-						source.src = res.mainUrl;
+						source.src = getResponseLink(res);
```
OrenMe: You can assign _url_ which you already have from previous call to the method

PR #2190, opened 2015-09-24
#SUP-5971 #comment disable start play btn ios #time 2h

GuyLivni:


itaykinnrot:
@ElizaSapir FYI on that

PR #2196, opened 2015-09-30
restore monitoring after pause resume for native controls.

mdale:


mdale:
fixes this test file: http://player.kaltura.com/modules/KalturaSupport/tests/NativeControls.html on chrome after you press play then pause then play on the native controls you get endless loading spinner; with fix monitoring is restored on native event play: http://kgit.html5video.org/pulls/2196/modules/KalturaSupport/tests/NativeControls.html

PR #2221, opened 2015-10-12
FEC-4219

amirch1:
enable user click to play when switching Youtube entries on iPhone


modules/ExternalPlayers/resources/mw.EmbedPlayerYouTube.js
```diff
@@ -410,6 +410,12 @@
 				callback();
 				if( mw.getConfig('autoPlay') || _this.isPlaylist){
 					if (mw.isMobileDevice()){
+						if (mw.isIphone()){
+							$(".largePlayBtn").hide();
```
OrenMe: I think we need a better way referring components instead of direct jquery CSS queries.

PR #2227, opened 2015-10-13
#SUP-6097 #comment unbind playerReady after tracks setup #time 2h

GuyLivni:


modules/KalturaSupport/resources/mw.ClosedCaptions.js
```diff
@@ -115,6 +115,7 @@
 						_this.destory();
 						_this.setupTextSources( function () {
 							_this.buildMenu( _this.textSources );
+							_this.unbind('playerReady');
```
itaykinnrot: @GuyLivni  please verify changemedia usecase

eitanavgil: @GuyLivni, @itaykinnrot 
This fix broke use case of  change-media + forcedCaptionLoaded. We MUST take it out or fix this issue on another approach. With this version of code any 2nd media is not dispatching forcedCaptionLoaded hence 3rd party plugins cannot know that a new caption file was loaded.

amirch1:
@GuyLivni - even when removing your fix, I still can't reproduce the issue on a local page. Lets discuss and see how we can reproduce it and fix it in another way.

amirch1:
@GuyLivni, @eitanavgil - reverted the fix and offer a new fix: https://github.com/kaltura/mwEmbed/pull/2286

PR #2244, opened 2015-10-16
Support drawing cuePoints types to scrubber plugin.

mdale:


mdale:
@itaykinnrot lets be sure to review / merge this is pretty small non-default change.

PR #2253, opened 2015-10-20
#SUP-6312 #comment fix custom string set #time 2h

GuyLivni:


modules/KalturaSupport/EntryResult.php
```diff
@@ -220,8 +220,11 @@ function getEntryResultFromApi(){
 		}
 
         $vars = $this->uiconf->playerConfig['vars'];
+		$playerConfig = $this->uiconf->getPlayerConfig();
 		if( is_array( $resultObject['contextData'] ) && isset( $resultObject['contextData']['code'] ) && $resultObject['contextData']['code'] == 'ENTRY_ID_NOT_FOUND' && !isset($vars['referenceId'])){
-			$this->error = 'No source video was found';
+			if (!isset($playerConfig['plugins']['strings']['mwe-embedplayer-missing-source'])){
+				$this->error = 'No source video was found';
```
itaykinnrot: @GuyLivni  and if its set ?

GuyLivni:
@itaykinnrot if its set, it will take the string message from: "mwe-embedplayer-missing-source".
Checked it works.
If the string is not set, or the string plugin is not enable it will say: No source video was found
If the string is set, it will select its value.

PR #2280, opened 2015-11-02
2.37 fec 4336

eitanavgil:


kWidget/onPagePlugins/omnitureOnPage/resources/omnitureOnPage.js
```diff
@@ -463,60 +478,80 @@ kWidget.addReadyCallback( function( playerId ){
...
+					//eval( this.getSCodeName() + '.Media.' + cmd + '("' + argSet.join('","') + '");');
+					// not working :(
+					//s.Media[cmd].apply( this, args );
+
+					if(_this.getConfig("overridePlayerName") != undefined ){
+						s.Media.playerName = String(_this.getConfig("overridePlayerName"));
+					}
+					if ( !s.inAd && cmd !== 'openAd') {
+						// re-evaluate mediaName even it it was already pushed to the stack
+						argSet[0] = _this.getMediaName();
```
itaykinnrot: @eitanavgil i think we should get the media name outside of the timeout in a temp var - since this can be called when ad is already complete for example - what do you think @eitanavgil

kWidget/onPagePlugins/omnitureOnPage/resources/omnitureOnPage.js
```diff
@@ -463,60 +478,80 @@ kWidget.addReadyCallback( function( playerId ){
...
+
+			var eventsTimeout = 0; //eventsTimeout is a new configuration, delay time in seconds
+			if (this.getConfig("eventsTimeout")){
+					eventsTimeout = this.getConfig("eventsTimeout")*1000;
+			}
+			_this = this;
+			var _args = args;
+
+
+			setTimeout(function(){
```
itaykinnrot: In order to make sure we are doing the same flow when timeout == 0 i would create a function for all the timeout and if timeout > 0 use set timeout if not just call the function as is.

eitanavgil: Done

kWidget/onPagePlugins/omnitureOnPage/resources/omnitureOnPage.js
eitanavgil: previousName is set only once on Open so only a media-open can change it. Since I don't see ads and timeout scenario now I will just add a configuration that I will turn on only when needed. What do you think of this?

kWidget/onPagePlugins/omnitureOnPage/resources/omnitureOnPage.js
```diff
@@ -463,61 +476,83 @@ kWidget.addReadyCallback( function( playerId ){
 			}
 		},
 		runMediaCommandWithArgs: function( args ){
+			var eventsTimeout = 0; //eventsTimeout is a new configuration, delay time in seconds
+			if (this.getConfig("eventsTimeout")){
+					eventsTimeout = this.getConfig("eventsTimeout")*1000;
+			}
+			_this = this;
```
itaykinnrot: @eitanavgil  missing var

PR #2283, opened 2015-11-04
add hideSource config to sourceSelector component

vladm3:


modules/KalturaSupport/components/sourceSelector.js
```diff
@@ -129,6 +130,15 @@
             if ( !this.handleAdaptiveBitrateAndContinue() )
                 return;
 
+            if (this.getConfig('hideSource')) {
+                // find and remove 'source' flavours (by tag)
+                $.each(sources.slice(0), function (sourceIndex, source) {
```
OrenMe: @vladm3 - you can use mediaElement removeSourceFlavor method that does the exact same, so we don't have duplicate code.

vladm3:
@OrenMe - thanks. Updated code to use removeSourceFlavor.

PR #2319, opened 2015-11-22
FEC-4433 Added new API service: getLicenseData.

noamtamim:
Returns a json object with license data (custom_data and signature) for each flavor of a given entry id. Also returns license server base url.

### DIFF
--- a/modules/KalturaSupport/apiServices/mweApiGetLicenseData.php
+++ b/modules/KalturaSupport/apiServices/mweApiGetLicenseData.php
@@ -15,50 +15,88 @@
 		},
 		"licenseServerBaseURL": "https://drm.example.com/license"
 	}
+	
+	OR, if there's an error:
+	{
+		"error": {
+			"message": "something is wrong"
+		}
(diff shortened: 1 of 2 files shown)


modules/KalturaSupport/apiServices/mweApiGetLicenseData.php
```diff
@@ -0,0 +1,85 @@
...
+	function getData() {
+		global $container;
+		$drmPluginData = null;
+		try {
+			// When EntryResult fails to parse some parameters (most notably ks), it fails when loaded.
+			$resultObject = $container['entry_result']->getResult();
+			$drmPluginData = (array)$resultObject["contextData"]->pluginData["KalturaDrmEntryContextPluginData"];
+			return $drmPluginData["flavorData"];
+		} catch (Exception $e) {
+			error_log("mweApiGetLicenseData: Failed to parse request: \n" . $e);
```
itaykinnrot: Please use our logger

PR #2320, opened 2015-11-23
#SUP-6602 #comment autoMute not supported on mobile #time 2h

GuyLivni:


itaykinnrot:
@ElizaSapir please review it - make sure its doesn't effect the mobile-sdk

PR #2324, opened 2015-11-24
Fix HLS playback on VOD

DBezemer:
Support HLS playback for VOD assets, also on desktop


itaykinnrot:
@einatr  please take a look

DBezemer:
@einatr @itaykinnrot any update? without this change you cannot view HLS VOD properly on desktop while the player with this change supports it.

einatr:
first of all, this branch has conflicts that you must resolve.
Besides, why do you wish to use streamerType 'applehttp'?
We already have 'hls' streamerType for hls urls.
In general we are trying to deprecate streamerType usage.

einatr:
we do support HLS VOD on desktop.

PR #2335, opened 2015-11-30
fix wrong vertical medalist height in mobile Safari (TR-608)

vladm3:


amirch1:
@vladm3 , @eitanavgil - can you send me a test page and device where I can see the problem? Using a timeout for layout updates is a bit hacky. Lets see if we can solve this another way.

vladm3:
@amirch1 - yes, I agree that timeout is pretty hacky. I'll try to find a better solution. I just thought that it's some known problem, because there is another setTimeout few lines above - https://github.com/vladm3/mwEmbed/blob/85c428a1dbd5bf09e309752a6e4ae211a550a61a/modules/KalturaSupport/resources/mw.KBaseMediaList.js#L313-L315 .

eitanavgil:
@vladm3 I lost track on this - do you know which player/widget caused this to happen? (maybe a TR specific bug)

vladm3:
@eitanavgil Yes, it was TR-608 bug

eitanavgil:
@vladm3 - looks like this caused a regression - vertical playlist on desktop now gets a vertical scrollbar even when not needed. Can you please review and suggest a new fix?

vladm3:
@eitanavgil - ok. I'll try to find another solution

vladm3:
I found the better solution for this, and it doesn't require core changes. The details are in the bug ticket.

PR #2366, opened 2015-12-10
#SUP-6647 #comment fix alert container with button #time 2h

GuyLivni:


OrenMe:
@GuyLivni - this will only be OK for certain situations.
The real issue is that we have a bottom bar and context that should be set differently.
The bottom bar with buttons should be fixed as a footer and the content should take all space between title and footer and get overflow so if it's too long it would be scrollable.
Also, we should have minimum size for alert box, so in mobile it won't be so small.

GuyLivni:
@OrenMe following our discussion, I have added scrolling option for overflowing text.
The max-height for the alert box was set to 44% and 40% following trial and error on mobile and desktop, makes the box a bit bigger, will have less effect now because the text is contained and scrollable within the alert container, but makes it look better.

PR #2424, opened 2016-01-03
SUP-6931

amirch1:
wait for changeMediaDone before requesting ads to prevent the media change callback destroying the ad manager in the Flash plugin


amirch1:
found a better fix

PR #2455, opened 2016-01-20
#SUP-6930 #comment display audio flavor in sourceSelector #time 2h

GuyLivni:


modules/KalturaSupport/components/sourceSelector.js
```diff
@@ -371,7 +373,9 @@
             if( source.getHeight() ){
                 title = this.getSourceSizeName( source ) + ' ';
             }
-            title += this.getSourceTitleBitrate(source);
+			if(!(source.getHeight() == 0)) {
```
amirch1: @GuyLivni - why not simply write: source.getHeight() !== 0

PR #2560, opened 2016-03-01
Webc 679

esakal:


esakal:
@OrenMe Can you please code review my changes? I already merged changes from 2.41 and it seems to be working.

Let me know if you have any questions. Note that this will be part of version 2.41 deployment.

@asafrob @kalturaguy FYI

modules/KalturaSupport/components/dualScreen/dualScreen.js
```diff
@@ -621,7 +637,133 @@
 					at: location[0]+location[1],
 					of: $( this.getPlayer().getInterface() )
 				});
-			}
+			},
+			getKClient: function () {
```
OrenMe: @esakal Is this being used somewhere? not seeing it in this PR.

modules/KalturaSupport/components/dualScreen/dualScreen.js
```diff
@@ -308,10 +323,12 @@
 
 					//Attach the primaryPlayerContainer to the primary display
 					var primaryDisplay = this.displays.getPrimary();
+					primaryPlayerContainer.attr('data-display-type','video');
 					primaryDisplay.attachView(primaryPlayerContainer);
 
 					//Attach the secondaryDisplay to the second display
 					var secondaryDisplay = this.displays.getSecondary();
+					secondaryPlayerContainer.attr('data-display-type','presentation');
```
OrenMe: vide/presentation display type is not necessarily a fixed state and will change with upcoming dual video feature, and this naming convention will make it quite confusing.
Please consult with @einatr before going along this path to make sure it coincides with her changes.

modules/KalturaSupport/components/dualScreen/dualScreen.js
```diff
@@ -621,7 +637,133 @@
...
+			/**
+			 * Gets all the cue points of type player view mode
+			 * @returns {Array}
+			 */
+			getPlayerViewModeCuePoints : function()
+			{
+				var _this = this;
+				var cuePoints = [];
+				if ( this.getPlayer().kCuePoints ) {
+					cuePoints = _this.getPlayer().kCuePoints.getCodeCuePointsByTag({tag : 'player-view-mode', sortDesc : false});
```
OrenMe: @esakal you are iterating the code cuepoints every clock tick to get code cuepoints by tag, this happens every 250ms and is very expensive. why not do this only when change happens, e.g. when you know for sure that new code cuepoint have arrived via API call.

OrenMe:
@esakal the requirment to add handling for live cuepoint handling for more then one type has changed the demands for kcuepoints service.
It was easy t reason about the logic when the requirement was only or one type, but the new demand would require refactoring.
I would suggest instead of handling each type in it's own array and with specific code paths to handle it in a map with hashes per type. This way we can reason about handling the cuepoints in pipe with out having to new where they are, e.g. different array names totally separate, but by type.
I would argue that most of the time we need same handling for most cuepoints.
Let's talk about this more so I can show you what I mean.

OrenMe:
@esakal I think separating all of the new logic of handing the view state via commands channel to it's own separate module that will be a component of the dual screen manager.
Think about in terms of your component is just another controller just as the control bar, all though they have some relation as they share the state of the view(need to manage in one place of course) and changes via command channel reflect in control bar UI of course.
In addition, it would be easier to manage in terms of separation of concerns if it resides in its own module.

modules/KalturaSupport/components/dualScreen/dualScreen.js
```diff
@@ -91,6 +93,17 @@
 								}).removeClass('firstScreen');
 								_this.hideDisplay();
 							}
+
+							// This plugin uses 'mw.KCuePoints' to retrieve que points, to persist the same conditions
+							// we listen to the same events as 'mw.KCuePoints'
+							$(_this.embedPlayer).bind(
```
OrenMe: @esakal if this is the same as kcuepoints why do we need an additional monitor? I would agree that seek operation reshuffles state and also kcuepoint service only updates the last closest cuepoint after a seek, but for all other use cases you can get it from cuepoint event.
If we don't fire code cuepoint event form kcuepoint it's because it wasnt required till now, but again, new requirements new code so we can remedy this in cuepoint service.

OrenMe:
@esakal The state transitions changes refactor - I would require some more design explanations, it seems to work for the control bar as before which is excellent, and I assume all other changes were made for the changes from cuepoints, but a lot of state management has moved inside the actual state machine instead of the state machine getting state names and mapping them out to actual explicit actions. 
Also, if I remember correctly the previous design also enabled defining the state transitions in a way that some transitions won't be possible(by design, for example mobile UX is different then desktop), is this still retained in this design?

modules/KalturaSupport/components/dualScreen/dualScreen.js
esakal: @OrenMe are you referring to the function 'getKClient'? This is leftover from initial implementation. I will remove it.

modules/KalturaSupport/components/dualScreen/dualScreen.js
esakal: @OrenMe agreed, this is why I'm not using that metadata in the final implementation. Will remove this also.

modules/KalturaSupport/components/dualScreen/dualScreen.js
esakal: @OrenMe I agree it is expensive, but since the dualscreen plugin is very complex we decided to use the same patterns/disciplines you are using to create a similiar flow with what we currently have.  

A. in file ['mw.kCuePoints.js' line 348](https://github.com/kaltura/mwEmbed/blob/webc-679/modules/KalturaSupport/resources/mw.KCuePoints.js#L348) you invoke function 'getNextCuePoint' every 250ms which does the same behavior. 
B. In file ['imageplayer.js' line 63](https://github.com/kaltura/mwEmbed/blob/webc-679/modules/KalturaSupport/components/dualScreen/imagePlayer.js#L63) if the cuepoint is not the one you are looking for you also invoke a loop against the list of cuepoints.

I can do enhancements to that logic like using a **priority queue** per cuepoint type which eliminate the need for re-looping the list every time, but I think this is something that the infrastructure should have handled. In order to **not affect** current image player behavior we decided to mimic the same behavior as done for the thumbs.

@OrenMe  @kalturaguy what do you think? Should I do the improvement for my part (only) but then the thumb code will still be doing the re-loop discipline and we will have different behavior in code?

esakal:
@OrenMe
All your comments about that service are true and I totally agree with you. in ideal world 'mw.kCuePoints.js' should have been a shared/infrasructure component that manages all the cuepoints without involving any business logic specific for the image player/thumb cuepoints. 

But currently this service has very specific behaviour. For example: * it manages  two separated internal lists of cuepoints (before I added the third one). * it relies on the assumption that only  cue point of type thumb will be handled. *it persist single cue point which make it quite hard to use it without refactoring it significantly. 

I can do the refactoring of the service to make it generic and decoupled but it will require more dev&qa effort. I added @kalturaguy to this comment to decide if he want to refactor this service as part of issue 679. If we prefer not to do that major refactoring at the moment then I think I managed to add the player view support with minimal impact on existing logics. 

Let's discuss together with @kalturaguy and decide how to continue with this service...

esakal:
@OrenMe regarding the component bar state machine. The major issue with this logic was that it was relying manged only the view type (PiP, SbS, single) and treat the position of the main/aux displays as a side story. In order to use the same state machine concept as you did I would need to refactor it to  support the complete list of view types (PiP-video, PiP-presentation,SbS-video, SbS-presentation, single-video, single-presentation). I don't see a significant advantage for it since the end result is the same.

We didn't loose any feature support since the states doesn't contain any information about the transition it self.

modules/KalturaSupport/components/dualScreen/imagePlayer.js
```diff
@@ -59,8 +59,11 @@
 						return false;
 					}
 				});
+
 				if (!cuePoint){
-					cuePoint = _this.getCurrentCuePoint();
+					// TODO [es]: TBD with Oren - I removed this since reached event will also trigger for code cue points and if I understood you correctly this workaround is redundant
+					// cuePoint = _this.getCurrentCuePoint();
```
esakal: @OrenMe I want to make sure that this workaround is not needed anymore since this function will be invoked also for the new code cue points and I don't want it to start searching for thumbs each time the new cue points is reached

modules/KalturaSupport/components/dualScreen/dualScreenControlBar.js
```diff
@@ -121,6 +110,63 @@
 			//Set tooltips
 			buttons.attr('data-show-tooltip', true);
 			this.embedPlayer.layoutBuilder.setupTooltip(buttons, "arrowTop");
+
+			// listen to state change and update buttons style accordingly
+			_this.bind( 'dualScreenStateChange', function(e, state){
```
esakal: @OrenMe this function is used to listen to changes in player view state and update the UI accordingly (disable/enable icons)

modules/KalturaSupport/resources/mw.KCuePoints.js
```diff
@@ -302,29 +303,67 @@
...
 			$(embedPlayer).bind(
 				"monitorEvent" + this.bindPostfix +
-					" seeked" + this.bindPostfix +
-					" onplay" + this.bindPostfix +
-					" KalturaSupport_ThumbCuePointsUpdated" + this.bindPostfix,
+				" seeked" + this.bindPostfix +
+				" onplay" + this.bindPostfix +
+				" KalturaSupport_ThumbCuePointsUpdated" + this.bindPostfix,
 				function (e) {
+					// TODO [es]: oren, we are we listening here for 'KalturaSupport_ThumbCuePointsUpdated'?
```
esakal: @OrenMe I suggest you to review this function from the source directly (without seeing the lines removed) - [link to source](https://github.com/kaltura/mwEmbed/blob/webc-679/modules/KalturaSupport/resources/mw.KCuePoints.js#L315)

modules/QnA/tests/QnA.html
```diff
@@ -22,12 +22,18 @@
 	kWidget.featureConfig({
 		'targetId': 'kaltura_player',
 		'wid': '_1802381',
-		'uiconf_id' : '30139111',
+		'uiconf_id' : '25906281',
```
esakal: oops :) should not be commited but since it is the test page I assume it is ok

esakal: @OrenMe If you will review the file changes you will see the I reverted all my changes in the state configuration so Einat will not be affected.

Most of the changes were made on the mw.cuePoints but I think that I managed to make specific modifications. 

Lets review it together tomorrow.

modules/KalturaSupport/components/dualScreen/dualScreen.js
```diff
@@ -51,11 +51,12 @@
 			fsmState: [],
 			screenShown: false,
 			currentScreenNameShown: "",
-
+			externalControlManager : null,
 			setup: function ( ) {
 				this.initConfig();
 				this.initDisplays();
 				this.initFSM();
+				this.initExternalControlManager();
```
esakal: @OrenMe I moved the initialization into the setup so it won't be affected from the changing media process, please confirm...

modules/KalturaSupport/components/dualScreen/dualScreenCuePointsManager.js
```diff
@@ -0,0 +1,191 @@
+(function (mw, $) {
+    "use strict";
+    mw.dualScreen = mw.dualScreen || {};
+
+    mw.dualScreen.CuePointsManager = function(name, bindPostfix, player, onCuePointsReached) {
```
esakal: @OrenMe this class manages the code cue points for the dual screen view

modules/KalturaSupport/components/dualScreen/dualScreenCuePointsManager.js
```diff
@@ -0,0 +1,191 @@
...
+
+        function getCodeCuePoints() {
+            return player.kCuePoints.getCodeCuePoints();
+        }
+
+        function initialize() {
+            nextPendingCuePointIndex = getNextCuePointIndex(0);
+
+            $player.bind('onChangeMedia', function () {
+                nextPendingCuePointIndex = getNextCuePointIndex(0);
```
esakal: @OrenMe does 'onChangeMedia' event is raised after the cuepoint service was updated with the new cuepoints?

modules/KalturaSupport/components/dualScreen/dualScreen.js
```diff
@@ -357,22 +366,28 @@
...
-						_this.fsm.consumeEvent( "switchView" );
-						if (_this.getPlayer().isAudio()){
-							_this.fsm.consumeEvent( "hide" );
-						}
-					}, 1000 );
 				} else {
 					showLoadingSlide();
 				}
+
+				var defaultDualScreenViewId = this.getConfig('defaultDualScreenViewId');
```
esakal: This logic replaces the previous handling of default view id. According to Renan we should not support scenario of audio only. 

We get the default view id and updates the view accordingly

esakal:
@OrenMe @kalturaguy - I updated the branch to reflect the modifications required by Oren.
Oren please review and let us know if you have any comments.


@asafrob FYI

OrenMe:
@esakal - approved, please merge after review with Einat.

PR #2564, opened 2016-03-02
Hls js

OrenMe:


modules/Hlsjs/Hlsjs.json
```diff
@@ -0,0 +1,10 @@
+{
+    "Hlsjs": {
+        "scripts": [
+            "resources/hls.min.js" ,
```
shacharz: Might be useful to state the version of the hls.js, since it updates frequently and have bug fixes. You'd like to know with what version you're running with.

PR #2572, opened 2016-03-06
FEC-5003,FEC-5010,FEC-5004

giladna:


itaykinnrot:
Confirmed on iOS/Android

PR #2588, opened 2016-03-10
hide mouse when in fullscreen mode

slash851:
If viewer is not moving the mouse - hide pointer after 3 second
To activate plugin add mouseHide.plugin : true to flashvars

Delay time can be changed from 3sec to any value via flashvar 'delayTime'


OrenMe:
Hi @slash851, thanks for the PR.
This kind of functionality would mainly be handled inside the layout manager.
You are welcomed to add it in your projects, and we will consider adding this into the player.

PR #2609, opened 2016-03-16
SUP-6501 Capturespace presentation ratio

ErezBouskila:
@OrenMe FYI


OrenMe:
@ErezBouskila looks good, let's merge, make sure QA cycle checks this on mobile devices as well

PR #2621, opened 2016-03-22
SUP-7442 added to the addbindings in the info plugin to load the plug…

ErezBouskila:
…in when the player is ready instead of the promise function


ErezBouskila:
@OrenMe FYI

PR #2625, opened 2016-03-23
FEC-5186,FEC-5187,FEC-5189 #comment fix back to live on hlsjs #time 5h

itaykinnrot:


modules/KalturaSupport/resources/mw.KBasePlugin.js
```diff
@@ -163,7 +163,7 @@ mw.KBasePlugin = Class.extend({
 
 		return false;
 	},
-	bind: function( eventName, callback ){
+	bind: function( eventName, callback , once ){
```
OrenMe: @itaykinnrot wouldn't it be better to add just bindOnce, more expressive then passing true in function call

modules/Hlsjs/resources/hlsjs.js
```diff
@@ -291,11 +294,13 @@
 			backToLive: function () {
 				var _this = this;
 				var vid = this.getPlayer().getPlayerElement();
-				vid.currentTime = vid.duration;
+				this.embedPlayer.goingBackToLive = true;
+				vid.currentTime = vid.duration - this.fragmentDuration * 3;
```
OrenMe: @itaykinnrot - maybe add default value 0 or 10 to this, in case frag change doesn't arrive yet, or doesn't update the value.

modules/KalturaSupport/components/live/liveStatus.js
```diff
@@ -58,20 +58,23 @@
...
-                }else {
-                    // live is off-synch
-                    _this.getPlayer().setLiveOffSynch(true);
-                    if ( _this.onAirStatus ) {
-                        _this.setOffSyncUI();
-                    }
-                    _this.prevIconClass = _this.unsyncIConClass;
-                }
-			});
+			this.bind('playing',function() {
```
OrenMe: @itaykinnrot -add _this.bindPostfix so listener is cleaned on destroy?

PR #2626, opened 2016-03-23
SUP-7442 add getScreen when player is ready to KBaseScreen 

ErezBouskila:
FYI @amirch1


amirch1:
@ErezBouskila - please re-open on 2.42-10-April

PR #2627, opened 2016-03-23
added check that the dual screen external manager is needed (only if …

esakal:
…live or has cuepoints)


esakal:
fix bug when merged into

PR #2641, opened 2016-03-29
SUP-7442 add to the addbindings in the info plugin to load the plugin when the player is ready

ErezBouskila:
FYI @amirch1


amirch1:
Waiting for a new pull request

PR #2683, opened 2016-04-12
SUP-6103 added loadThumbnailsWithReferrer flashvar and based on flash…

omridevk:
…var load thumbnail with referrer or not.

Moved the comment to the JIRA.

Please let me know what you think. 

@OrenMe


omridevk:
Will add a test page.

PR #2695, opened 2016-04-18
SUP-7960 letting the adLoader know that content is complete so cuepoi…

omridevk:
…nts can request new ads

FYI @amirch1

### DIFF
--- a/modules/DoubleClick/resources/mw.DoubleClick.js
+++ b/modules/DoubleClick/resources/mw.DoubleClick.js
@@ -873,8 +873,14 @@
 			// Make sure the  this.getAdDisplayContainer() is created as part of the initial ad request:
 			this.getAdDisplayContainer();
 			this.hideAdContainer(false);
+			// Ad loader already exists no need to load it again.
+			// Enough to signal it that the content is complete.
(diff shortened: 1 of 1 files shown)


omridevk:
need to check still why ended listener is not fired. 

Maybe this commit is not needed (we are doing contentComplete() there).

omridevk:
hmm not sure if this is a viable solution since the contentComplete suppose to let the SDK knows that player finished playing and that the SDK can continue with the pre-proll. not sure if that is what we want.

omridevk:
maybe need to check if there's a postRoll, if there isn't one, we can signal contentComplete.

omridevk:
More info:

https://developers.google.com/interactive-media-ads/docs/sdks/html5/faq#8

omridevk:
Also
Maybe use:
_this.embedPlayer.setKalturaConfig('doubleClick', "adTagUrl", cuePointWrapper.cuePoint.sourceUrl);

to set the adTagUrl when handling cuePoints. this way, we can also use an adTagUrl while tracking cuePoint. to make sure this is fully supported and working, only need to dismiss the AdsLoader after the cuePoint finished and load the original adTagUrl and in the background, if there's a post-roll we can reload the original adTagUrl

PR #2702, opened 2016-04-19
Fix HLS AES source selection for DRM allowed sources

OrenMe:


modules/MultiDRM/MultiDRM.loader.js
```diff
@@ -140,7 +140,7 @@
 			var hlsSource = sources.filter( function ( source ) {
 				return ( source.mimeType === "application/vnd.apple.mpegurl" );
 			});
-			drmSources.push(hlsSource);
+			drmSources.concat(hlsSource);
 		}
```
amirch1: @OrenMe - so hlsSource might contain more than one source? In that case, maybe call it hlsSources. but feel free to merge if you don't feel like changing the variable name...

OrenMe: No, but filter returns an array so in order to merge both we require concat and not push.

Sent from Outlook Mobile<https://aka.ms/qtex0l>




On Tue, Apr 19, 2016 at 9:49 PM -0700, "Amir Chervinsky" <notifications@github.com<mailto:notifications@github.com>> wrote:


In modules/MultiDRM/MultiDRM.loader.js<https://github.com/kaltura/mwEmbed/pull/2702#discussion_r60350159>:

> @@ -140,7 +140,7 @@
>                       var hlsSource = sources.filter( function ( source ) {
>                               return ( source.mimeType === "application/vnd.apple.mpegurl" );
>                       });
> -                     drmSources.push(hlsSource);
> +                     drmSources.concat(hlsSource);
>               }


@OrenMe<https://github.com/OrenMe> - so hlsSource might contain more than one source? In that case, maybe call it hlsSources. but feel free to merge if you don't feel like changing the variable name...

-

amirch1: Thats exactly what I meant. Never mind - merging, we can change the name later if we want...

PR #2713, opened 2016-04-26
TR-976 Omniture calls aren't sent after hitting 'Replay' button

vladm3:
fixes:
* TR-976 Omniture calls aren't sent after hitting 'Replay' button
* remove trailing spaces


amirch1:
A different solution was used to solve this issue: https://github.com/kaltura/mwEmbed/pull/2715

PR #2717, opened 2016-04-26
Fix error in DoubleClick skip button

jpluijmers:
When an ad has a shorter duration than the skipOffset, the hideSkip button method will not be called, because all ad functionality gets broken down after the ad finishes. Leaving the skip button permanently in place. This update will not show the button at all, since it is not necessary in those circumstances.

### DIFF
--- a/modules/DoubleClick/resources/mw.DoubleClick.js
+++ b/modules/DoubleClick/resources/mw.DoubleClick.js
@@ -716,6 +716,11 @@
 			}
 		},
 		showSkipBtn: function(){
+			var offset = this.embedPlayer.getKalturaConfig( 'skipBtn', 'skipOffset' ) || 15;
+			if ( this.duration < offset ){
+				return
+			}
+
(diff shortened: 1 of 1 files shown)


amirch1:
@jpluijmers - thanks. The hideSkip method is always called when the ad is complete. For [Javascript](https://github.com/kaltura/mwEmbed/blob/master/modules/DoubleClick/resources/mw.DoubleClick.js#L1162) SDK and for [Flash](https://github.com/kaltura/mwEmbed/blob/master/modules/DoubleClick/resources/mw.DoubleClick.js#L1285) SDK.
Seems that the ad label text is not always removed. I've added a fix for that issue here: https://github.com/kaltura/mwEmbed/commit/56d27eaf11170922d9b6f60e2ed3ec65d1fb8ec9
Please let me know if you encounter any issues after this fix. If not - please close this pull request.

jessp01:
@jpluijmers - any update on this one? Thanks.

jessp01:
@jpluijmers - closing due to lack of feedback. Please open if the issue is not resolved as per @amirch1's response.

PR #2723, opened 2016-04-27
webc-691 - updating qna plugin to support new producer settings updat…

ainvoner:
…e action

### DIFF
--- a/modules/QnA/resources/qnaService.js
+++ b/modules/QnA/resources/qnaService.js
@@ -584,12 +584,12 @@ DAL for Q&A Module
                     request['filter:updatedAtGreaterThanOrEqual'] = lastUpdatedAt;
                 }
 
-                // build list code cue point request
+                // build list code cue point request for qna settings (checking both new and old tags for BC)
                 var codeCuePointListRequest = {
                     'service': 'cuepoint_cuepoint',
                     'action': 'list',
                     'filter:entryIdEqual': entryId,
-                    'filter:tagsLike':'WEBCASTSTATETAG',
+                    'filter:tagsMultiLikeOr':'player-qna-settings-update,WEBCASTSTATETAG',
                     'filter:cuePointTypeEqual': 'codeCuePoint.Code',
                     'filter:orderBy': '-createdAt',
                     'pager:pageSize': 1,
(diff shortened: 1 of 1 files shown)


modules/QnA/resources/qnaService.js
```diff
@@ -584,12 +584,12 @@ DAL for Q&A Module
...
                 }
 
-                // build list code cue point request
+                // build list code cue point request for qna settings (checking both new and old tags for BC)
                 var codeCuePointListRequest = {
                     'service': 'cuepoint_cuepoint',
                     'action': 'list',
                     'filter:entryIdEqual': entryId,
-                    'filter:tagsLike':'WEBCASTSTATETAG',
+                    'filter:tagsMultiLikeOr':'player-qna-settings-update,WEBCASTSTATETAG',
```
esakal: we removed ** WEBCASTSTATETAG** tag and will not send it anymore, you can remove it. If you are saw other places that still use it let me know

modules/QnA/resources/qnaService.js
```diff
@@ -671,18 +671,34 @@ DAL for Q&A Module
...
-
-                        if (cuePoint.tags && cuePoint.tags.indexOf("WEBCASTSTATETAG")>=0) {
-                            try {
-                                var webCastingState=JSON.parse(cuePoint.code);
-                                disableModule=!webCastingState["QnA"];
-                                announcementOnly=!disableModule && webCastingState["AnnouncementsOnly"];
+                        //for BC supporting both new and old QnA settings cue points
+                        if(cuePoint.tags) {
+                            //old QnA settings cue point convention //todo [sa] remove after releasing new producer version
+                            if (cuePoint.tags.indexOf("WEBCASTSTATETAG") >= 0) {
```
esakal: same here. this code is obselete

modules/QnA/resources/qnaService.js
ainvoner: @esakal since we need to support the old convention (up until both player and webcast app will get upgraded) I'm supporting both cases

modules/QnA/resources/qnaService.js
ainvoner: @esakal will be removed only after both player and webcasting app will get updated

esakal: Good catch :)

esakal:
@ainvoner I merged your changes into webc-691 and created a shared pull request https://github.com/kaltura/mwEmbed/pull/2728

PR #2728, opened 2016-05-02
Webc 691

esakal:


esakal:
@ainvoner fyi

esakal:
@einatr Those changes are relevant for [slides-upload feature](https://kaltura.atlassian.net/browse/WEBC-690) that was moved from the KMS into the webcast producer app. This added the following use-cases:
- Starting web-cast without any slide
- Removing selected slide (showing blank image)

-----

To make things simple we only remove the content of the image player when needed. This means that:
- Changing the view to parent-only (video only) is done by sending a separated cue point (and not by the dual screen plugin). 
- We don't disable/hide the change view buttons when we are not showing any thumb.

-----

This branch was merged from version 44 (June version).

@kalturaguy, @OrenMe please coordinate this task

OrenMe:
Hi @esakal pay attention you are pull requesting against 2.43, please open the PR against 2.44.
Also, u changed the version to 2.44.rc11 which doesn't exist. please don't change this value, we are taking care of this so always resolve from the repository value.
Regarding naming conventions, in general please don't change it to specific use cases. I'm referring to "manager" cue points. AFAIK the BE doesn't have such name and unless you can promise no one will ever use code cue points I rather we keep the original name.
I understand that the above change is due to the fact you are now requiring to mix cuepoint types for the feature, but this is binding the kcuepoints to specific use cases and naming patterns of an external applicative demands which is not desired.
Please find a way to manage such complexities outside the core, or you are welcomed to suggest another way if you want.
Thanks.

PR #2734, opened 2016-05-04
added grace period to ignore minor time shifting

esakal:


esakal:
@einatr Can you please take a look on this bug fix? It is quite straight forward

PR #2737, opened 2016-05-04
SUP-7515 set the volume after each ad loaded event

omridevk:
will move to relevant branch once it is opened.


omridevk:
FYI @itaykinnrot

PR #2740, opened 2016-05-05
forcing default view to be no-parent (slide only) when broadcasting a…

esakal:
…udio only


OrenMe:
@ErezBouskila  - please link to ticket and verify fix

esakal:
@ErezBouskila The fix for this issue was merged as part of other branch as seen here:
https://github.com/kaltura/mwEmbed/blob/2.44-5-June/modules/KalturaSupport/components/dualScreen/dualScreen.js#L410

Can I close this pull request? Did you test the fix in version 2.44?

PR #2741, opened 2016-05-08
Webc 691 (merged from 2.44-5-June)

esakal:


esakal:
@OrenMe thanks for you feedback, This pull request was created following your [comment on this pull request](https://github.com/kaltura/mwEmbed/pull/2728#issuecomment-217230469).  

Regarding the version change I assume it happened since my pull request was against version 43 while I merged the right branch of 44. I don't see it in this pull request anymore.

This feature major modification was done actually on the [ImagePlayer.js](https://github.com/kaltura/mwEmbed/pull/2741/files#diff-04db33308a8426901b475f045e63430c) file since i was forced to create a single flow to manage reached cuepoints.

The change in kCuePoints is minor and was done mainly to remove potential confusion in the future. It is basically only variable naming modification. If you recall this component serves that cuepoints to the dual screen only as a bypass to existing logic (for backward compatible reasons). While adding it in version 42 we called the variable **Code**CuePoints which was relevant back then. While adding this feature we extended it to serve also **Thumb** cue points. So I just extended the feature from previous sprint and I didn't want to use the value 'dualScreen' inside the kCuePoints although this is its purpose at the moment.

I would appreciate your inputs about what I wrote when you will be available.

modules/KalturaSupport/resources/mw.KCuePoints.js
```diff
@@ -236,16 +238,19 @@
...
 				var thumbNewCuePoints = [];
-				var codeNewCuePoints = [];
+				var managerNewCuePoints = [];
 				//Only add new cuepoints
 				$.each(rawCuePoints, function (id, rawCuePoint) {
 					if (!_this.associativeCuePoints[rawCuePoint.id]) {
 						_this.associativeCuePoints[rawCuePoint.id] = rawCuePoint;
 
-						if (rawCuePoint.cuePointType === 'codeCuePoint.Code')
+						if (rawCuePoint.cuePointType === 'codeCuePoint.Code' || rawCuePoint.cuePointType === 'thumbCuePoint.Thumb')
```
OrenMe: @esakal the new manager cuepoints type is a composite type which holds both thumb and code cuepoints and it serves only one use case which is not core.
I would like to avoid such coupling in core components as possible.

esakal:
@OrenMe We have completed a regression test of this branch by our QA. Can you please merge it into the player branch of version 2.44?

@asafrob @kalturaguy FYI

esakal:
@OrenMe hold on... I need to merge first...

esakal:
@OrenMe I'm ready for merge... can you please handle it?
Thanks

PR #2748, opened 2016-05-09
Changed the CURL TIMEOUT to 1 second

supermasita:
Changed the 10 seconds TIMEOUT to 1 second. This setting can consume all the PHP slots on a busy enviroment when an adserver is slow, and bring the platform down. This happened for real on a big deployment.

It is a super simple change but critical (IMHO)

Thanks!

### DIFF
--- a/simplePhpXMLProxy.php
+++ b/simplePhpXMLProxy.php
@@ -216,8 +216,8 @@ if ( !$url ) {
 	// Always follow redirects: 
 	curl_setopt( $ch, CURLOPT_AUTOREFERER, true );
 		
-	// Add a total curl execute timeout of 10 seconds: 
-	curl_setopt( $ch, CURLOPT_TIMEOUT, 10 );
+	// Add a total curl execute timeout of 1 second: 
+	curl_setopt( $ch, CURLOPT_TIMEOUT, 1 );
 	
 	if ( strtolower($_SERVER['REQUEST_METHOD']) == 'post' ) {
 		curl_setopt( $ch, CURLOPT_POST, true );


kaltura-hooks:
Hi @supermasita,

supermasita:
Is this doable?

OrenMe:
@supermasita are you saying that it take up to 10 seconds to get a reply from the adServer? can you share why when requests can take so long?

supermasita:
Hi @OrenMe !
Yes, there are time when the ad servers can take longer then 10 seconds. The usual reasons are faulty ad servers, network issues or sub standard ad services. 

There is no way to have control over what ad server performance, but anyways (IMHO) 10 seconds is too much to wait for a (non critical) service to answer, making it really easy to create a "snowball effect".

Thanks!

OrenMe:
@supermasita I understand your claim.
The issue is that this might be critical, cause usually ads playback means revenue stream for AVOD.

I would rather let the user have a fluctuating delay, where most of the times the ad server will respond in time and capped at a maximum level. 
If ads requests start dropping after more than 10 seconds and this affects the server, and as you stated, this is probably due to faulty ad server, then the user should contact his/hers ad server provider to handle the service.

supermasita:
@OrenMe 

If the **whole platform** is down due to this issue, then no video is played and no revenue is generated. Also, in the particular case this was found, only one partner had issues with their ad server, but the brought the whole platform down - all partners unusable. 

If you believe that 1 second is too low, please consider something higher that you feel comfortable with (3 seconds?). 

Thanks,

OrenMe:
@supermasita thanks for the suggestion, please let me check the issue further and update.
By the way, are we talking about a SaaS deployment or this is an onPrem issue?

supermasita:
@OrenMe this was a OnPrem issue - but I guess it could theoretically happen in SaaS if you get the same conditions.

OrenMe:
@supermasita thanks.

OrenMe:
Hi @supermasita, this issue is not affecting the SaaS environment. I assume this is related to the way it is setup.
If this is affecting your onPrem feel free editing it to your needs.

Thanks you for reporting this issue.

PR #2756, opened 2016-05-13
Make comscore compatible with studio

DBezemer:


OrenMe:
Handled in https://github.com/kaltura/mwEmbed/pull/3046

PR #2762, opened 2016-05-17
#SUP-7410 #comment Removed second playback speed increment on iPad  #…

ErezBouskila:
…time 3h


ErezBouskila:
@amirch1 I only checked if there is serverSpeedPlayback in the onRemovePlayerSpinner.

PR #2786, opened 2016-05-25
FEC-5601 #comment Fix hlsjs changeMedia API

OrenMe:


modules/Hlsjs/resources/hlsjs.js
```diff
@@ -473,10 +473,10 @@
 			/**
 			 * Override player callback after changing media
 			 */
-			changeMediaCallback: function(){
-				this.getPlayer().play();
```
amirch1: @OrenMe - thinking of it, we must verify playlist auto continue when using hlsjs after removing the changeMediaCallback call. Might get broken...

OrenMe: This change retuned the changeMediaCallback flow, and just overrides the actual vid.src switch with the hlsjs logic of switching.
All other logic should remains the same.
In any case, I verified it and it's OK. Thanks!

PR #2790, opened 2016-05-26
sup-7804 - handle replay when the vod starts without a slide and late…

esakal:
…r has a slide. the slide remains when replying

### DIFF
--- a/modules/KalturaSupport/components/dualScreen/imagePlayer.js
+++ b/modules/KalturaSupport/components/dualScreen/imagePlayer.js
@@ -75,19 +75,21 @@
 				this.cuePointsManager = null;
 			}
 		},
-		cuePointsReached : function(context)
-		{
+		cuePointsReached : function(context) {
 			var cuePoints = context.filter({tags: ['remove-selected-thumb']});
 			cuePoints = cuePoints.concat(context.filter({types: this.slidesCuePointTypes}));
 
 			cuePoints.sort(function (a, b) {
-				return  (b.startTime - a.startTime);
+				return (b.startTime - a.startTime);
 			});
 
 			var mostUpdatedCuePointToHandle = cuePoints.length > 0 ? cuePoints[0] : null; // since we ordered the relevant cue points descending - the first cue point is the most updated
 
 			if (mostUpdatedCuePointToHandle) {
 				this.sync(mostUpdatedCuePointToHandle)
+				return true;
+			} else {
+				return false;
 			}
 		},
 		canRender: function () {
@@ -100,6 +102,20 @@
 				)
 			);
 		},
+		syncByReachedCuePoints : function()
+		{
+			var _this = this;
+			if (_this.cuePointsManager) {
+				var cuePointsReachedResult = _this.cuePointsManager.getCuePointsReached();
+				if (cuePointsReachedResult)
+				{
+					return _this.cuePointsReached(cuePointsReachedResult);
+
+				}
+			}
+
+			return false;
+		},
 		addBinding: function(){
 			var _this = this;
 			this.bind( 'playerReady', function (  ) {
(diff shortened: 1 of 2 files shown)


modules/KalturaSupport/components/dualScreen/imagePlayer.js
```diff
@@ -75,19 +75,21 @@
...
 			cuePoints.sort(function (a, b) {
-				return  (b.startTime - a.startTime);
+				return (b.startTime - a.startTime);
 			});
 
 			var mostUpdatedCuePointToHandle = cuePoints.length > 0 ? cuePoints[0] : null; // since we ordered the relevant cue points descending - the first cue point is the most updated
 
 			if (mostUpdatedCuePointToHandle) {
 				this.sync(mostUpdatedCuePointToHandle)
+				return true;
```
esakal: this is only indication if we had cue points to handle

modules/KalturaSupport/components/dualScreen/imagePlayer.js
```diff
@@ -100,6 +102,23 @@
 				)
 			);
 		},
+		syncByReachedCuePoints : function()
```
esakal: This invoke a sync by cue points. If no sync was done it indicate that we should not show any image

modules/KalturaSupport/components/dualScreen/imagePlayer.js
```diff
@@ -100,6 +102,23 @@
...
+		{
+			var _this = this;
+			if (_this.cuePointsManager) {
+				var cuePointsReachedResult = _this.cuePointsManager.getCuePointsReached();
+				if (cuePointsReachedResult)
+				{
+					if (!_this.cuePointsReached(cuePointsReachedResult))
+					{
+						// when user seek/press play we need to handle scenario that no relevant cue points has reached and thus we need to clear the image shown.
+						_this.sync(null);
```
esakal: If no sync was done it indicate that we should not show any image

modules/KalturaSupport/components/dualScreen/imagePlayer.js
```diff
@@ -100,6 +102,19 @@
...
 			);
 		},
+		syncByReachedCuePoints : function()
+		{
+			var _this = this;
+			if (_this.cuePointsManager) {
+				var cuePointsReachedResult = _this.cuePointsManager.getCuePointsReached();
+				if (!_this.cuePointsReached(cuePointsReachedResult)) {
+					// when user seek/press play we need to handle scenario that no relevant cue points has reached and thus we need to clear the image shown.
+					_this.sync(null);
```
esakal: If we didn't handle any cuepoints, it indicates that we should remove any visible image

modules/KalturaSupport/components/dualScreen/imagePlayer.js
```diff
@@ -108,6 +123,7 @@
 
 			this.bind( 'onplay', function () {
 				_this.loadAdditionalAssets();
+				_this.syncByReachedCuePoints();
```
esakal: we would like to invoke explicit sync when user start playing or perform a seek (I verified with @amirch1 that 'onplay' event will be triggered if seeking while playing)

esakal:
@OrenMe I pushed my fix for sup-7804, please review and merge to rc version so the QA will be able to test it.

modules/KalturaSupport/components/dualScreen/imagePlayer.js
OrenMe: @esakal what happens if you seek while paused? playback state is retained on seek, e.g. if you were playing and preformed a seek you would keep playing after seek, if you were seeked while paused seek will happen but you would still be paused afterwards.

modules/KalturaSupport/components/dualScreen/imagePlayer.js
```diff
@@ -100,14 +102,36 @@
...
 		addBinding: function(){
 			var _this = this;
 			this.bind( 'playerReady', function (  ) {
 				_this.initializeCuePointsManager();
 			});
 
+			this.bind('seeked',function()
+			{
+				// Checking if we are pausing, if we do then we need to handle sync from 'seeked' event. otherwise the 'onplay' event will handle the sync
+				if (!_this.getPlayer().isPlaying())
```
esakal: @OrenMe I added the following logic to handle 'seeked' event only when pausing.

esakal:
@OrenMe please review and let me know what you think... thanks :)

PR #2811, opened 2016-06-06
example fix for SUPPS-686 (TR-885)

vladm3:
added forceCompatMode flashVar to passthrough IE compat meta tag

### DIFF
--- a/modules/KalturaSupport/kalturaIframeClass.php
+++ b/modules/KalturaSupport/kalturaIframeClass.php
@@ -579,11 +579,13 @@ class kalturaIframeClass {
 	}
 
 	function outputCompatibilityTags(){
-		$flashvars = $this->request->getFlashVars();
-		if (isset($flashvars) &&
-				isset($flashvars['forceCompatMode']) &&
-				!empty($flashvars['forceCompatMode'])){
-			return '<meta http-equiv="X-UA-Compatible" content="' . $flashvars['forceCompatMode'] . '"/>';
+		if (isset($this->request) && !empty($this->request)) {
+			$flashvars = $this->request->getFlashVars();
+			if (isset($flashvars) &&
+					isset($flashvars['forceCompatMode']) &&
+					!empty($flashvars['forceCompatMode'])){
+				return '<meta http-equiv="X-UA-Compatible" content="' . $flashvars['forceCompatMode'] . '"/>';
+			}
 		}
 
 		return '';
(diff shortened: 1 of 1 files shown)


modules/KalturaSupport/kalturaIframeClass.php
```diff
@@ -578,6 +578,17 @@ private function getEntryTitle(){
 		return "Kaltura Embed Player iFrame";
 	}
 
+	function outputCompatibilityTags(){
+		$flashvars = $this->request->getFlashVars();
```
itaykinnrot: please add check to the request (not null) just to be on the safe side @vladm3 
after that we can merge it - Thanks!

vladm3:
@itaykinnrot Added the check to the request.

modules/KalturaSupport/kalturaIframeClass.php
```diff
@@ -578,6 +578,19 @@ private function getEntryTitle(){
 		return "Kaltura Embed Player iFrame";
 	}
 
+	function outputCompatibilityTags(){
+		if (isset($this->request) && !empty($this->request)) {
```
OrenMe: @vladm3 - any reason for the added security check here?
In general we have an IoC that injects the request into the kalturaIframeClass, so unless you saw it's necessary please remove the redundant check.

modules/KalturaSupport/kalturaIframeClass.php
```diff
@@ -578,6 +578,19 @@ private function getEntryTitle(){
...
+			$flashvars = $this->request->getFlashVars();
+			if (isset($flashvars) &&
```
OrenMe: @vladm3 the same goes for here, getFlashVars method always return an array(or a default value if you supply one), and if you look closely you will see that a few lines above where you call this method("outputCompatibilityTags") we also call "$this->request->getFlashVars()" so it is available to you also from there.

modules/KalturaSupport/kalturaIframeClass.php
```diff
@@ -578,6 +578,19 @@ private function getEntryTitle(){
...
+					isset($flashvars['forceCompatMode']) &&
```
OrenMe: in general "!empty" is the same as checking if it isset as well, and even if it's not set you won't get a warning from PHP on empty check on an undeclared var. AFAIK.

eitanavgil:
@OrenMe  can you please take this with your team?

vladm3:
@OrenMe thanks for the clues. I made some code rearranging and removed all redundant checks.

PR #2815, opened 2016-06-08
SUP-8286 update poster right after the quiz is loaded

omridevk:


omridevk:
FYI @mrkfdr 
Can you please review?

alon1980:
@omridevk  i thought we said we will use the other fix

PR #2852, opened 2016-06-21
FEC-5710

amirch1:
prevent native apps from transforming numbers, dates, emails and addresses to corresponding application links


modules/KalturaSupport/kalturaIframeClass.php
```diff
@@ -1284,6 +1284,10 @@ function getIFramePageOutput( ){
 		echo '<script src="' . $wgRemoteWebInspector . '"></script>';
 	 } ?>
 	<link href='//fonts.googleapis.com/css?family=Lato:400,700' rel='stylesheet' type='text/css'>
+    <meta name="format-detection" content="telephone=no">
```
OrenMe: I suggest we enter this only when needed, e.g. user agent detection?

PR #2872, opened 2016-06-30
add loop attribute handling (TR-1202)

vladm3:


eitanavgil:
@OrenMe 
@vladm3  had found an issue with loop - can you please review this pull?

PR #2876, opened 2016-07-03
SUP-8636 disable the hover controls in KMS and not in the quiz plugin…

omridevk:
… and using flashvars in deployment quiz player

@OrenMe 
Can't find the right time to disable hovering, so easier solution would be to add it via FlashVar in KMS and as a uiVar for new deployed players in KMS


omridevk:
FYI @mrkfdr

PR #2883, opened 2016-07-07
#SUP-6410 Related plugin clickUrl modification to open in new tab

ErezBouskila:
@OrenMe

### DIFF
--- a/modules/KalturaSupport/components/related/related.js
+++ b/modules/KalturaSupport/components/related/related.js
@@ -385,18 +385,21 @@
 			data["autoSelected"] = (auto === true);
 			this.getPlayer().sendNotification('relatedVideoSelect', data);
 
-			if(this.getConfig('clickUrl')){
+			if( this.getConfig('clickUrl') ){
+
 				try {
 					if( this.getConfig( 'openInNewTab' ) === true ) {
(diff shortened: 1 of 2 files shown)


modules/KalturaSupport/components/related/related.js
```diff
@@ -386,11 +387,15 @@
 
 			if(this.getConfig('clickUrl')){
 				try {
+					if( this.getConfig( 'openInNewTab' ) === true ) {
+						window.open(this.getConfig('clickUrl'), '_blank');
+						return;
```
OrenMe: @ErezBouskila better use if/else and return once.

omridevk: @ErezBouskila , better yet, use Guard Clauses and replace the condition logic, so you can have one return at the top. More info:
http://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html

omridevk: maybe it won't fit here and you will have to follow Oren's instructions.

ErezBouskila: @OrenMe modified the fix

PR #2886, opened 2016-07-11
Chromecast ui/x

itaykinnrot:
* make sure the player ui is visible
* add default configuration to the chromecast UI based on mobile skin
* change the largePlayBtn to TV mode (show pause when paused)
* add option to pass different uiconf to the chromecast plugin

### DIFF
--- a/skins/kdark/css/layout.css
+++ b/skins/kdark/css/layout.css
@@ -195,7 +195,7 @@ body {
     font-size: 1.5em;
 }
 
-.mobileSkin.pause-state .titleLabel{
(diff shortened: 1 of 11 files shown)


itaykinnrot:
@OrenMe  @amirch1  FYI

skins/kdark/css/layout.css
```diff
@@ -195,6 +195,11 @@ body {
     font-size: 1.5em;
 }
```
amirch1: @itaykinnrot - this affects all mobile devices and not only Chromecast causing the title to appear on pause which is not the required behaviour.

PR #2887, opened 2016-07-11
#SUP-8596 if bumper plugin has postSequence restore player normally

omridevk:


omridevk:
not sure if this viable solution, since not sure it is good that doubleClick will be aware of the bumper plugin.

omridevk:
close PR. mistakes were done.

PR #2908, opened 2016-07-26
FEC-5623

vladm3:


eitanavgil:
@OrenMe  please review

eitanavgil:
@OrenMe can you please review this?

OrenMe:
@vladm3 , @eitanavgil  - this is very costly in terms of DOM access.
Better to solve this by saving state when action actually happens.

eitanavgil:
@vladm3 can you suggest a solution that is state-based?

vladm3:
@OrenMe , @eitanavgil 
I guess there's no some kind of universal event which is fired for all drop-down based plugins when menu opens or closes. So, in order to track the state of drop-downs I'd have to add logic to every plugin with drop-down and it doesn't feel like a rational solution.
Another way to do this is to bind to the click event like this:
`controlBarContainer.on('click', '<some drop down selector>')` but this is also nearly as costly as the current solution.
Maybe I'm missing something but I'll continue digging into this and try to find the optimal solution.

vladm3:
@OrenMe , @eitanavgil 
Added drop down menus caching, so we don't need to make DOM lookup every time.

eitanavgil:
@OrenMe  @dan-ziv  
Can you review this and merge or comment please?

OrenMe:
@eitanavgil seems quite well, but we must have performance test here.
@vladm3 please preform some tests for this, using [performance markers](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now).
We must make sure such changes don't affect run time performance.
If you have any questions please reach out personally to me.

eitanavgil:
@vladm3 do you know how to use performance marker?

eitanavgil:
@OrenMe as this is a core issue - can't we use QA resources to test this?

OrenMe:
@eitanavgil QA cannot test such a change in any manner, it's black box to them.
This is DOM access change and need to be verified by dev.

vladm3:
@OrenMe, @eitanavgil 
I've run performance tests on hidePlayerControls method - it's the only method of changed which can be ran a lot of times. The results are:

Added code ([lines 724-729](https://github.com/vladm3/mwEmbed/blob/0b6af21ee9eb9d08136e022e501f2b0ebf587d3a/modules/EmbedPlayer/resources/mw.PlayerLayoutBuilder.js#L724-L729)) avg time: **0.03ms**
Already existing code ([lines 730-735](https://github.com/vladm3/mwEmbed/blob/0b6af21ee9eb9d08136e022e501f2b0ebf587d3a/modules/EmbedPlayer/resources/mw.PlayerLayoutBuilder.js#L730-L735)) avg time: **1.45ms**

So, basically, the introduced changes slow down the hidePlayerControls method by 0.03ms in average.

eitanavgil:
@OrenMe  can we process this now?

OrenMe:
@vladm3 - please fix merge conflicts so we can merge it.
@eitanavgil - FYI.

PR #2919, opened 2016-07-31
SUP-8683_sort_captions_alphabetically

ErezBouskila:
@OrenMe

### DIFF
--- a/modules/KalturaSupport/resources/mw.ClosedCaptions.js
+++ b/modules/KalturaSupport/resources/mw.ClosedCaptions.js
@@ -443,13 +443,11 @@
 					// Done adding source issue callback
 					mw.log( 'mw.ClosedCaptions:: loadCaptionsURLsFromApi> total captions count: ' + captions.length );
 					// Check if we need to sort captions array Alphabetically
-					if( _this.embedPlayer.getKalturaConfig( "closedCaptions", "sortCaptionsAlphabetically" )) {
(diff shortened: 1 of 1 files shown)


modules/KalturaSupport/resources/mw.ClosedCaptions.js
```diff
@@ -441,10 +442,24 @@
 					_this.captionURLs = captionsURLs;
 					// Done adding source issue callback
 					mw.log( 'mw.ClosedCaptions:: loadCaptionsURLsFromApi> total captions count: ' + captions.length );
-					callback( captions );
+					// Check if we need to sort captions array Alphabetically
+					if( _this.embedPlayer.getKalturaConfig( "closedCaptions", "sortCaptionsAlphabetically" )) {
```
OrenMe: @ErezBouskila you should just use `this.getConfig("sortCaptionsAlphabetically")`

modules/KalturaSupport/resources/mw.ClosedCaptions.js
```diff
@@ -441,10 +442,24 @@
...
+						captions = _this.sortByKey( captions, 'language' );
+						callback( captions );
+					}
+					else {
+						callback( captions );
```
OrenMe: Instead of `if(){}else{}` just use `if`, and always call `callback( captions );`

PR #2922, opened 2016-08-01
Webc 714 add lock view mode

ainvoner:


esakal:
Hi @ainvoner,
In file `` there are many placed that check if the plugin should show/enable the control bars.

I know that you inherit this logic (not something that you wrote) but since you are touching it I think you should create a function on the plugin and call it everywhere:

```
{
   canManipulateControlViews : function()
   {
      return !_this.disabled && _this.controlBar && !_this.getPlayer().isAudio() && !_this.isViewModeLocked;
   }
}
```

This will fix places where they didn't validate one of the conditions.

esakal:
@ainvoner 
Also in `dualScreen.js` (I cannot put a comment inside the file since you didn't modify that part):

```
this.bind("onChangeMedia", function(){
   ...
   this.isViewModeLocked = false; // this is a must because when changing media inside a channel playlist we need to remove the lock before we handle the next entry
   ...
});
```

esakal:
@ainvoner,  **IMPORTANT**
From which branch did you checkout? You should have done it from [webc-734](https://github.com/kaltura/mwembed/tree/webc-734)

Because it contains all our changes for the poll.

Problem : You modified file 'modules/KalturaSupport/components/dualScreen/dualScreenExternalControlManager.js' which not exists anymore. 

I think you should merge from that branch and copy your changes in this file manually.

modules/KalturaSupport/components/dualScreen/dualScreenExternalControlManager.js
```diff
@@ -104,6 +104,11 @@
                 }
 
             },
+            setViewModeLockState : function (viewModeState)
+            {
+                mw.log("dualscreenExternalControlManager.setViewModeLockState(): Changing player view mode state to '" + viewModeState);
```
esakal: use `this.log` instead of `mw.log`

PR #2956, opened 2016-08-19
Fem 726

ElizaSapir:


ElizaSapir:
@OrenMe , @amirch1  I made this small change only for getting the play method called on my side.
This is not a solution BUT you can see that when I ignore these to "return false" play method is called on my side and replay supported.

If you nee me for debugging let me know.

Thanks!

cc @itaykinnrot

PR #2974, opened 2016-08-24
- Fixed a problem where the {{ns_st_vo}} label was reporting the valu…

joezeitouny:
…e zero when the player volume is set to anything between 0 and 99.

- Fixed a problem where no measurements are being sent after resuming playback following a pause in live streams.
- Fixed a problem where in some cases a {{pause}} measurement was sent when the video ends instead of an {{end}} measurement.
- Fixed a problem where the plugin wasn't sending out an {{end}} measurement to notify of ending playback when the attached player instance is destroyed.


OrenMe:
@joezeitouny  - can you please open PR against dev branch 2.48

PR #3015, opened 2016-09-12
SUP-8883

amirch1:
update buttons icon color and shadow


OrenMe:
@amirch1 why are we using important here?

PR #3019, opened 2016-09-12
Enhance force loading captions

eitanavgil:


modules/KalturaSupport/resources/mw.ClosedCaptions.js
```diff
@@ -494,10 +498,13 @@
 			// Return a "textSource" object:
 			return new mw.TextSource( embedSource, _this.embedPlayer );
 		},
-		forceLoadLanguage: function(){
-			var lang = this.getConfig('forceLoadLanguage');
+		forceLoadLanguage: function(language){
+			var lang = this.getConfig('forceLoadLanguage') ? this.getConfig('forceLoadLanguage') : language;
```
OrenMe: @eitanavgil why not just use the already existing `'forceLoadLanguage'` option to accomplish this?

modules/KalturaSupport/resources/mw.ClosedCaptions.js
```diff
@@ -494,10 +498,13 @@
...
 			var source = this.selectSourceByLangKey( lang );
-			// Found caption
+			//fallback - failed to find a matching asset - fallback to first found if exist
+			if(source == null && this.textSources.length){
```
OrenMe: @eitanavgil this will present a new behviour which might surprise anyone not expecting it.
Although it might be reasonable I prefer we add this from application side and not introduce new behviour here which is not expected.

modules/KalturaSupport/resources/mw.ClosedCaptions.js
eitanavgil: @OrenMe  
Because I want this function to also serve the new API from line 156 so that external plugins will be able to load a specific language. 
Example - external plugin gets "fr" from metadata and is interested to load French, it will send a notification "forceLoadCaptions" with "fr" as argument.

modules/KalturaSupport/resources/mw.ClosedCaptions.js
eitanavgil: @OrenMe  
We thought of it too, but this fallback functionality will also support setting forceLoadLanguage=true which will load any language and not fail.

OrenMe: @eitanavgil - following our discussion can this be closed?

PR #3025, opened 2016-09-14
SdkBridgeChanges

OrenMe:
Using jQuery triggerHandler instead of trigger command.
Trigger emits an event but in addition also looks for the event name on
the event target prototype and if it finds it then it calls it.
So if play event arrives it will emit play event, but will also call
play method on it, resulting in double play.


OrenMe:
Already done on another PR

PR #3038, opened 2016-09-21
Clear timout of skip btn when changing media

dan-ziv:
FEC-6116

### DIFF
--- a/modules/DoubleClick/resources/mw.DoubleClick.js
+++ b/modules/DoubleClick/resources/mw.DoubleClick.js
@@ -1680,7 +1680,7 @@
 		destroy:function(){
 			// remove any old bindings:
 			var _this = this;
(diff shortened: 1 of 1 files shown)


modules/DoubleClick/resources/mw.DoubleClick.js
```diff
@@ -1680,6 +1680,7 @@
 		destroy:function(){
 			// remove any old bindings:
 			var _this = this;
+			clearTimeout(_this.skipTimeoutId);
 			if ( this.adTagUrl || this.currentAdSlotType === "postroll" ){
```
amirch1: @dan-ziv - please use something like:
`			if( this.skipTimeoutId !== null ){
				clearTimeout(this.skipTimeoutId);
				this.skipTimeoutId = null;
			}`

PR #3067, opened 2016-10-11
SUP-8163 Revert the fix for the removal of the black pixel poster since video …

ErezBouskila:
…tag play button on Android is shown

@amirch1  @giladna


OrenMe:
Already done on another PR.

PR #3080, opened 2016-10-20
Fixed a problem where a TypeError was fired after playback switches t…

joezeitouny:
…o the next playback item in a playlist.


OrenMe:
@joezeitouny is this needed for next release? can't see diff on github due to size so I guess this is a big change.

joezeitouny:
Yes this is needed to be released as soon as possible. Since the code changes contain fixes for critical issues that affect the way we report analytics data in our plugin.

OrenMe:
@joezeitouny - merged, will be available on 11/7 on production.

PR #3081, opened 2016-10-20
FEC-6200 #comment prevent endless loop when destroying

yairans:


amirch1:
@yairans - please define the 'isdestroy' flag as a plugin property set to false and reset it to false on playerReady. @OrenMe -FYI

PR #3082, opened 2016-10-21
SUP-8939 fix volume slider in IE11 iframe fullscreen and add user's o…

ErezBouskila:
…peration system detection

@OrenMe


ErezBouskila:
change to get and remove os that are not supported

PR #3083, opened 2016-10-23
SUP-9408 'updateLayout' infinite loop when CC layout is below

ErezBouskila:
@OrenMe In the end did not need to add a flag since the CC change size when open/close full screen (External CC as well).

Revert the first part of this commit - https://github.com/kaltura/mwEmbed/pull/3002

### DIFF
--- a/modules/KalturaSupport/resources/mw.ClosedCaptions.js
+++ b/modules/KalturaSupport/resources/mw.ClosedCaptions.js
@@ -193,7 +193,7 @@
 				}
 			});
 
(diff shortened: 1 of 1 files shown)


OrenMe:
@ErezBouskila the actual events are called before the fullscreen API is called so the dimensions that are calculated when the event is triggered are still before the fullscreen state change and hence sizes are not yet updated.

omridevk:
Closing. no longer relevant, we went with a different solution.

PR #3097, opened 2016-10-27
Sup 8825 fix i os flavor transition

ErezBouskila:
@OrenMe  @yairans made the change to load only when using iPad


itaykinnrot:
@ErezBouskila can you please explain the change?

PR #3099, opened 2016-10-30
peer5 plugin update

mrbar42:
This is a change needed as part of the integration simplification.
(supersedes older PRs)


OrenMe:
@mrbar42 - merged

mrbar42:
@OrenMe can you point me to the correct staging server url?
or its only possible through releases and pre-release (i.e. tags)?

OrenMe:
@mrbar42 - in general it is available only via tags

PR #3100, opened 2016-10-30
SUP-9408 do not trigger updateLayout in CC setup

omridevk:
@OrenMe 

Let me know what you think, this will make sure that we are not entering an infinite loop, i.e we are listening to the updateLayout event and triggering updateTextSize which then causes layout again and so on.

Let me know if you think this is a viable solutions. 

Thanks.


omridevk:
closing. not relevant anymore, we have chosen a different solution (flag indicating that event was triggered by the same function).

PR #3101, opened 2016-10-30
SUP-9408 another solution for updatelayout infinite loop inside close…

omridevk:
…dCaptions plugin

@OrenMe 

This the second solution I had in mind, let me know what you think.

(I think it is fine to pass data around if it maybe undefined, let me know if you think otherwise and I'll check for default values)


omridevk:
@OrenMe 
Per our discussion, moved to local flag as suggested. please review.

PR #3115, opened 2016-11-05
Re-add DRM mime type selection to native SDK (#3088)

giladna:


giladna:
Oren,

I am afraid that actually this fix is not 100% OK 
since in case of OTT when they ask for Tablet Main (HLS format) and multiDRM is ON
so no source will come....

please advice

giladna:
Seems that in this Case just need to :disable Multi DRM since we want HLS

PR #3116, opened 2016-11-06
SUP-9337 add ks to serveWebVTT api request if available

omridevk:
@OrenMe 

Let me know if this will suffice (tested and seems to work) to solve the KS issue.

thanks.


omridevk:
@OrenMe 

Maybe it will be better to expose another FlashVars, something like loadWebVTTWithKs and check if it is set, don't see a reason to pass the KS in another cases, for cache reason and more.

omridevk:
Closing. move to branch 2.51

PR #3127, opened 2016-11-10
SUP-8926 dont minimize or restore view with ads if entry is audio

omridevk:


omridevk:
discussed with @esakal , he approved this PR. @OrenMe  Can you please merge?
FYI @ErezBouskila

PR #3128, opened 2016-11-13
SUP-6852 remove num of clips because we cannot distinguish between ru…

omridevk:
…le based playlist and regardless we dont want to do extra API calls to figure out the number of clips


omridevk:
Since we don't want to do extra API calls to figure out the number of clips, it is best to just not do it at all for all playlists regardless if it is a rule based playlist(where the content is XML and not CSV of entries) or normal playlist.

omridevk:
@OrenMe 
Can you please review?

omridevk:
wrong JIRA number, closing moving to new PR

PR #3129, opened 2016-11-13
SUP-7353 removing number of clips from multiple playlist view all but…

omridevk:
… the current playlist

### DIFF
--- a/modules/DoubleClick/resources/mw.DoubleClick.js
+++ b/modules/DoubleClick/resources/mw.DoubleClick.js
@@ -39,7 +39,7 @@
 
 		// flag for using a chromeless player - move control to KDP DoubleClick plugin
 		isChromeless: false,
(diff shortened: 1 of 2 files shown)


omridevk:
Since we don't want to do extra API calls to figure out the number of clips, it is best to just not do it at all for all playlists regardless if it is a rule based playlist(where the content is XML and not CSV of entries) or normal playlist.

omridevk:
@OrenMe 
Can you please review (same PR as the other one with a new branch name with the correct JIRA)

omridevk:
I have removed a string attribute:
'mwe-embedplayer-videos'

not sure what will happen for those that uses it and if we should alert them of this change somehow?

OrenMe:
@omridevk please fix conflict with 2.51 so we can merge

PR #3134, opened 2016-11-16
updated the docs to use URL params instead of forward slashes

zoharbabin:
there was a mistake in how autoembed was documented. this pull fixes this issue.


OrenMe:
@zoharbabin please open PR against 2.51 branch

PR #3143, opened 2016-11-21
#opf-6596 fix control bar spacing

GuyLivni:


modules/EmbedPlayer/resources/mw.PlayerLayoutBuilder.js
```diff
@@ -335,19 +335,26 @@
...
+
+            // Check if specific container was provided otherwise go over all containers and update their components
+            if(container){
                 _this.updateContainerCompsByAvailableSpace(
-                    _this.getInterface().find('.' + containerId)
+                    _this.getInterface().find('.' + container)
                 );
-            });
+            } else {
+                $.each(this.layoutContainers, function (containerId, components) {
```
omridevk: why is the else needed here? why not return true after getInterface and then can get rid of the extra indentation

PR #3176, opened 2016-12-18
FEC-6308 remove role attribute

gonenradai:


OrenMe:
@gonenradai - shouldn't there be a replacement? or we can just drop it at all?

gonenradai:
@OrenMe 
I discussed this with the vendor and removing this attribute is enough in this case.
especially since this is in the flash object (which is going to be mostly irrelevant soon).

PR #3178, opened 2016-12-18
add memcache support + make iframe cache use the cache provider

itaykinnrot:


modules/KalturaSupport/kalturaIframeClass.php
```diff
@@ -516,8 +518,9 @@ private function getMwEmbedStartInline(){
 			
 		// check for cached startup:
 		if( !$wgEnableScriptDebug){
-			if( is_file( $cachePath ) ){
-				return file_get_contents( $cachePath );
+			$content = $this->cache->get( $cachePath );
```
OrenMe: Till now we used to create the _$cachePath_ with the _$wgScriptCacheDirectory_ (a few lines above this) and operate on it directly.
If we move to using the cache adaptors then I think we need to remove it from _$cachePath_ as we init the adaptor with _ wgScriptCacheDirectory_ in [KalturaCommon.php](https://github.com/kaltura/mwEmbed/blob/master/modules/KalturaSupport/KalturaCommon.php#L59)
The same goes to the other changes where we create the file cache path.

itaykinnrot:
Fixed @erankor , @OrenMe  review

PR #3180, opened 2016-12-19
FEC-6159 #comment relying on parent monitor handler and init clockSta…

yairans:
…rtTime only in 'firtsPlay'


modules/EmbedPlayer/resources/mw.EmbedPlayerImageOverlay.js
```diff
@@ -46,6 +46,14 @@
 			}
 		},
 
+		setup: function (readyCallback) {
+			this.bindHelper('onplay' + this.bindPostfix, function(){
+				// Reset clock time for load
+				this.clockStartTime = new Date().getTime();
```
OrenMe: Won't this be triggered on each play?

PR #3188, opened 2016-12-20
SUP-8514 use offset to shift the array of thumbassets

omridevk:
@OrenMe 
Option # 2 that I had in mind.


OrenMe:
Solved in #3190

PR #3209, opened 2017-01-15
Universal analytics

jpluijmers:
Hi,

This is a publication of the UniversalAnalytics plugin I developed, based of the GoogleAnalytics plugin. The entire plugin / event structure was left intact since this has been battle tested for many years. 

Only two things should have changed:
- analytics.js instead of ga.js interface
- support UniversalAnalyticsEvent, allows for other plugins to trigger UA events without altering the UA plugin.

Let me know what you think!


jpluijmers:
I have no clue why those old commits are being shown. The changes are all mine though.

modules/UniversalAnalytics/resources/mw.UniversalAnalytics.js
```diff
@@ -0,0 +1,385 @@
...
+			return this.embedPlayer.getKalturaConfig('universalAnalytics', attr);
+		},
+
+		init: function (embedPlayer, callback) {
+            var _this = this;
+            this.embedPlayer = embedPlayer;
+            // Unbind any existing bindings
+            this.embedPlayer.unbindHelper(_this.bindPostFix);
+
+            eval(ANALYTICS_PAYLOAD);
```
itaykinnrot: We don't allow eval - i prefer we'll just load their JS code from a url

jpluijmers:
@itaykinnrot 

Replaced with <script> inject. Also added a credits.MD, hope you don't mind.

DBezemer:
@OrenMe @itaykinnrot gently reminding you that this pull requets is still open. We have some customers asking whether we can support Universal Analytics instead of the legacy analytics, and it will allow for better events reporting (for example using the new analytics events as well)

OrenMe:
stale PR, new GA analytics is implemented in playkit player

PR #3216, opened 2017-01-16
Add "checkboxes" for chromecast, airplay and native callout plugins.

jpluijmers:
Allow for clients to enable these plugins from the studio.


OrenMe:
Hi @jpluijmers.
Chromecast was recently added.
We are not planing to add native callout at the moment as it is a deprecated feature.
AirPlay needs to be upgraded and only then can be added.

Thanks,

Oren M.

PR #3222, opened 2017-01-18
Fix sourceSelector for streamerType http

jpluijmers:
The sourceSelector will not list any flavours except auto. This is because the handleAdaptiveBitrateAndContinue method returns false. Removed the HTTP case for the moment, since auto is irrelevant for "HTTP".

The HLS and Dash cases rebuild the menu anyways with a different streamerType.


OrenMe:
@jpluijmers this has been addressed. You can test it on latest player.

OrenMe:
Closing due to inactivity.
If you feel this is still relevant please reopen.

PR #3223, opened 2017-01-18
Sup 10006 use p tag instead of br for captions

omridevk:


ErezBouskila:
@OrenMe

modules/TimedText/resources/mw.TextSource.js
```diff
@@ -438,7 +438,9 @@
 					}
 					if( caption ){
 						// concatenate text lines to html text
-						caption['content'] = s.slice(2).join("<br>");
+						caption['content'] = s.slice(2).reduce(function(ac, current) {
```
OrenMe: IE8 doesn't support reduce method

omridevk: can't we add the Polyfill for reduce?

OrenMe: @omridevk is this an array slice and concat? I believe you can achieve the same result with array concat woth seperator

OrenMe:
Solved in https://github.com/kaltura/mwEmbed/pull/3280

PR #3273, opened 2017-02-12
KMS-13602 - accessibility

GuyLivni:


eitanavgil:
@GuyLivni  looks good - once we get the about span VS aria we will finalize this pull.

Issue #3274, opened 2017-02-13
Chromecast doesn't show when player URL is in iframe


slash851:
When I use the kaltura url directly chromecast plugin is displayed on the control bar container:

http://www.kaltura.com/index.php/extwidget/preview/partner_id/1822451/uiconf_id/38373291/entry_id/1_omx6h6un/embed/auto

But when I place that url inside of the iframe on my page chromecast disappear

http://karol.html-5.me/kaltura_iframe.html

The same is happening when I place emed code (for example dynamic embed) to my page - chromecast works. But when I put that page in iframe on other domain it disappear.

kaltura-hooks:
Hi @slash851,

Thank for you reporting an issue and helping improve Kaltura!

To get the fastest response time, and help the maintainers review and test your reported issues or suggestions, please ensure that your issue includes the following (please comment with more info if you have not included all this info in your original issue):
 - Is the issue you're experiencing consistent and across platforms? or does it only happens on certain conditions?
    please provide as much details as possible.
 - Which Kaltura deployment you're using: Kaltura SaaS, or self-hosted?
	If self hosted, are you using the RPM or deb install?
 - Packages installed.
	When using RPM, paste the output for:
```
	# rpm -qa \"kaltura*\"
```	
	For deb based systems:
```	
	# dpkg -l \"kaltura-*\"
```
 - If running a self hosted ENV - provide the MySQL server version used
 - If running a self hosted ENV - is this a single all in 1 server or a cluster?
 - If running a self hosted ENV, while making the problematic request, run:
```
	# tail -f /opt/kaltura/log/*.log /opt/kaltura/log/batch/*.log | grep -A 1 -B 1 --color \"ERR:\|PHP\|trace\|CRIT\|\[error\]\"
```
   and paste the output.
 - When relevant, provide any screenshots or screen recordings showing the issue you're experiencing.

For general troubleshooting see:
https://github.com/kaltura/platform-install-packages/blob/Jupiter-10.13.0/doc/kaltura-packages-faq.md#troubleshooting-help

If you only have a general question rather than a bug report, please close this issue and post at:
http://forum.kaltura.org

Thank you in advance,

slash851:
Looks like attaching the cast_sender.js via jQuery instead of appending it through the kWidget.appendScriptUrl method solve the problem

Works for me now.

OrenMe:
@slash851 issue fixed: #3861
Thanks for reporting and sorry for the delay!

PR #3282, opened 2017-02-16
#3274 chromecast doesnt show when player url is in iframe

slash851:
Issue #3274  
Attach the code using jQuery instead of kWidget.appendScriptUrl
since within nested iframes kWidged method was causing CORS issue


OrenMe:
@slash851 please add a test page showing the issue.
Thanks.

OrenMe:
@slash851 issue fixed: #3861
Thanks for reporting and sorry for the delay!

PR #3283, opened 2017-02-16
Fix default captions selection for IOS

GuyLivni:


modules/KalturaSupport/resources/mw.ClosedCaptions.js
```diff
@@ -598,6 +606,32 @@
 				this.selectedSource = this.textSources[0];
 			}
 		},
+		selectDefaultIosTrack: function (defaultLangKey) {
+			var _this = this;
```
OrenMe: @GuyLivni all native HTML5 logic should be in the native.js implementation.
And in general FYI - you can use `this.embedPlayer.getPlayerElement()` instead of `_this.embedPlayer.getVideoHolder().find("video")[0]`

modules/KalturaSupport/resources/mw.ClosedCaptions.js
```diff
@@ -565,17 +565,25 @@
 				if( defaultLangKey == 'None' ){
 					return ;
 				}
+				if ( mw.isIOS() ) {
```
OrenMe: @GuyLivni not sure we want this for ipad as well, cause in ipad we might show our own SRT rendered captions.

GuyLivni: Sure, will add check for iPad

PR #3288, opened 2017-02-26
Add default audio by RFC-693 lang setting

OrenMe:
Can be configured via audioSelector plugin with “defaultLang”:
LANG_NAME, where LANG_NAME is an RFC-693-3 string
FEC-6493


modules/KalturaSupport/components/audioSelector.js
```diff
@@ -114,7 +115,27 @@
...
+					return false;
+				});
+				if (defaultStreams.length > 0) {
+					this.log("found default language by key: " + defaultStreams[0].language);
+					return defaultStreams[0];
+				}
+				this.log("warn - language key "+ defaultLangName +" not found, get default language by index: " + this.streams[defaultLangIndex].language);
+				return this.streams[defaultLangIndex];
+			}
+			this.log("found default language by index: " + this.streams[defaultLangIndex].language);
```
yairans: is `this.streams[defaultLangIndex]` can't be null?

OrenMe: U r right, added protection.

PR #3304, opened 2017-03-12
SUP-9622 Fix source selector logic

ErezBouskila:


ErezBouskila:
@OrenMe When using Silverlight it shows all the sources as expected so did not need to make any changes there.

PR #3307, opened 2017-03-15
SUP-9113 Limit resizing dimensions for dualScreen

ErezBouskila:
@OrenMe as discussed to limit of the draggable area will be handled by Vlad


OrenMe:
Merged

PR #3310, opened 2017-03-15
SUP-9932 playbackRateSelector key code not working on Firefox

ErezBouskila:
@OrenMe only FF has different hotkeys, tested on all browsers.


modules/KalturaSupport/resources/uiConfComponents/playbackRateSelector.js
```diff
@@ -126,6 +126,24 @@
 		},
 		addKeyboardShortcuts: function( addKeyCallback ){
 			var _this = this;
+			
+			//FF has different key code for the +,-,=
+			if ( mw.isFirefox() ) {
```
OrenMe: Why duplicate the entire code when you only need to set the key code value?

PR #3311, opened 2017-03-19
SUP-9095 IE8 square playhead

ErezBouskila:
@OrenMe the issues is fixed once removing the importent, I checked with all other browsers to see if the removal has any effecte and all is working as expected.


OrenMe:
@ErezBouskila add check to run this only on IE8

OrenMe:
@ErezBouskila you are repeating a lot of code there.
Why not do 
`var cssString = "style","background-color: " + value;`
`if (!mw.isIE8()){`
`    cssString += " !important";`
`}`
`$(".mwPlayerContainer:not(.mobileSkin)").find(".playHead").attr(cssString );`

PR #3312, opened 2017-03-19
CC queues - migrate up next notification

dan-ziv:


OrenMe:
@dan-ziv please fix conflicts

PR #3318, opened 2017-03-21
fix dependency bug for dualVideo

eitanavgil:


eitanavgil:
@OrenMe  see this fix - worked well for me in localhost.

PR #3321, opened 2017-03-23
Fix pause video after post roll ends in case no video sibling

dan-ziv:
FEC-6537


OrenMe:
Also cherry-pick to 2.55

PR #3324, opened 2017-03-23
SUP-10644 move ad container div to the back as it interferes with pla…

ErezBouskila:
…yer big play button on mobile


OrenMe:
Need to handle on PS UI with z-indexing

PR #3327, opened 2017-03-26
Receiver change media fixes

dan-ziv:


modules/KalturaSupport/resources/mw.KDPMapping.js
```diff
@@ -1254,6 +1254,15 @@
...
 
+                        if ( typeof notificationData.proxyData === 'string' ) {
+                            try {
+                                notificationData.proxyData = JSON.parse( notificationData.proxyData );
+                            } catch ( e ) {
+                                mw.log("Unable to parse proxyData", notificationData.proxyData);
+                                embedPlayer.changeMediaStarted = false;
+                                return;
+                            }
+                        }
```
OrenMe: Need to make sure we do all rest of operations(nullify the managers and update poster)  in change media and only then return.
Also add in the log output that change media is canceled.

PR #3341, opened 2017-04-06
add new flag for same domain on iOS for 360 video and re-embed the pl…

itaykinnrot:
…ayer with the same domain of the video


OrenMe:
Closing due to inactivity.

PR #3347, opened 2017-04-12
SUP-9742 Player icons sometimes not loaded

ErezBouskila:


ErezBouskila:
@OrenMe we still have the issue of reproducing, as you know we cannot always reproduce the issue so I am not sure my fix is correct here.

Also we need to take into consideration the KMS, I am not sure it will fix the issue there.

ErezBouskila:
@OrenMe added the flag

OrenMe:
@ErezBouskila change name to `ForceLayoutRedraw `

PR #3366, opened 2017-04-24
new DVR branch against 2.56

eitanavgil:


modules/KalturaSupport/components/chapters/chapters.js
```diff
@@ -981,8 +991,13 @@
 				if (!this.getPlayer().canAutoPlay() && this.getPlayer().firstPlay){
 					this.getPlayer().sendNotification('doPlay');
 				}
-				// see to start time and play ( +.1 to avoid highlight of prev chapter )
-				this.getPlayer().sendNotification('doSeek', ( this.mediaList[mediaIndex].startTime ) + 0.1);
+				if (this.embedPlayer.isDVR()) {
+					var seekTo = this.mediaList[mediaIndex].startTime - this.embedPlayer.evaluate('{mediaProxy.entry.firstBroadcast}') + this.embedPlayer.kalturaPlayerMetaData.firstBroadcast - this.embedPlayer.kalturaPlayerMetaData.currentBroadcastStartTime;
```
kalturaguy: 1.mediaProxy.entry.firstBroadcast. currentBroadcastStartTime is  not something you should use...
- You don't know which DC is serving you, the currentBroadcastStartTime is not per DC, and the two DC are not in sync..
- It's "deprecatd", and we plan to remove this one day, it's not a value you should depend 
2. the DVR window can be smaller than the session so even if we update currectly the currentBroadcastStartTime you will have a problem here

modules/QnA/resources/qnaService.js
```diff
@@ -164,7 +164,10 @@
             return this.cuePoint().text;
         };
 
-        this.isAnswerOnAirActive = function(playerTime) {
+        this.isAnswerOnAirActive = function(playerTime,player) {
+            if(player && player.embedPlayer.isDVR()){
```
kalturaguy: this looks like a bug player.embedPlayer.currentTime contains the absolute time, not relative one

eitanavgil: @OrenMe 
We are talking about live+dvr 
player.embedPlayer.currentTime is holding a relative time **in seconds** and not absolute timestamp. Is this a bug? Should it be holding the absolute time? (and if we switch this - wouldn't it break many other components?)

PR #3368, opened 2017-05-01
Analytics - Fix code property duplicate incrementation

dan-ziv:
FEC-6552


OrenMe:
@dan-ziv isn't the issue with another flag? `firstPlayDone` flag? did you check it before adding another flag?

PR #3373, opened 2017-05-07
Run Chromecast only on Chrome browser

dan-ziv:


OrenMe:
@dan-ziv please commit to 2.58

PR #3376, opened 2017-05-11
SUP-10985 #comment support native HLS text tracks

yairans:


modules/EmbedPlayer/resources/mw.EmbedPlayerNative.js
```diff
@@ -1556,6 +1571,36 @@
...
 		},
+		onSwitchTextTrack: function (event, data) {
+			var vid = this.getPlayerElement();
+			var textTracks = vid.textTracks;
+			if (textTracks && textTracks.length) {
+				if (!data) {
+					var activeSubtitle = this.getActiveSubtitle(textTracks);
+					if (activeSubtitle) {
+						activeSubtitle.mode = 'hidden';
+						this.log('onSwitchTextTrack disable subtitles');
```
itaykinnrot: we have this.log here? or mw.log?

modules/EmbedPlayer/resources/mw.EmbedPlayerNative.js
```diff
@@ -1452,20 +1453,34 @@
...
+			                    'srclang': textTrack.label,
+			                    'label': textTrack.label,
+			                    'title': textTrack.label,
+			                    'id': textTrack.id,
+			                    'index': i
+		                    });
+	                    }
+	                    textTrack.mode = 'hidden';
+                    }
+                    if (textTracksData.languages.length) {
```
itaykinnrot: Check that the object is not null before checking length

yairans: not necessary. i just created it some lines above with languages as empty array (#1462).

modules/EmbedPlayer/resources/mw.EmbedPlayerNative.js
yairans: both of them. for simple string used this, for string and object used mw

PR #3382, opened 2017-05-15
FEC-6597 CVAA styling applied to embedded captions

ErezBouskila:


ErezBouskila:
Dose not support FireFox:
https://developer.mozilla.org/en-US/Apps/Fundamentals/Audio_and_video_delivery/Adding_captions_and_subtitles_to_HTML5_video

Under Consideration in Microsoft Edge on Desktop:
https://developer.microsoft.com/en-us/microsoft-edge/platform/status/shadowdom/

modules/KalturaSupport/resources/mw.ClosedCaptions.js
```diff
@@ -905,6 +911,30 @@
 			style["background"] = "none";
 			return style;
 		},
+		getCvaaCssForEmbedded: function(cvaaCss) {
```
omridevk: I would make a function to map between those properties, something like this:
https://github.com/jquery/jquery/blob/master/src/core.js#L293

itaykinnrot: And change the name of the function

modules/KalturaSupport/resources/mw.ClosedCaptions.js
```diff
@@ -29,7 +29,8 @@
 			'smartContainerCloseEvent': 'changedClosedCaptions',
 			"forceWebVTT": false, // force using webvtt on-the-fly. only for kalturaAPI captions
 			"enableOptionsMenu": false,
-			"sortCaptionsAlphabetically": false
+			"sortCaptionsAlphabetically": false,
+			"cvaaOnEmbeddedCaptions": false
```
itaykinnrot: @ErezBouskila  lets change the name here - as it not related only to cvaa

PR #3388, opened 2017-05-23
SUP-10279 Related module player functionality

ErezBouskila:


OrenMe:
@ErezBouskila please add detailed explanation for why the PR is solving your issue, it is not understood in any way.

ErezBouskila:
@OrenMe Before clicking play the related plugin is disabled (Grayed out) as it dose not have any related videos, if you open the share or info plugin, after you close the screen/menu of the plugins the related plugin will be active. I added to the preHideScreen of each plugin to exclude the related and thus it will stay disabled. After the first play the related is enable.

modules/KalturaSupport/components/related/related.js
```diff
@@ -266,12 +266,11 @@
...
 		},
 		isValidResult: function( data ){
-			// Check if we got error
+			// Check if we got error and if entry list is empty disable the button
 			if( !data || data.length === 0	||( data.code && data.message )	){
 				var errMsg = data.message ? data.message : 'No related items were found.';
 				this.log('Error getting related items: ' + errMsg );
-				this.updateTooltip(gM('mwe-embedplayer-related-errMsg'));
-				this.onDisable();
+				this.getBtn().hide();
```
OrenMe: If you change media and that media has related, will the button show?

ErezBouskila: This will not happen as you well know, with the related if you select the "Related to entries" option it works with tags but if there are no tags it presents a list from the sphinx DB. If you select the "Playlist ID" or "Entries list" then it will always display that list for the entire playlist.

The only way for the related to not display entries is if you put an invalid entry id in the "Entries list".

ErezBouskila: So with a playlist the related is disabled for the entire list of entries and will not come back if the condition is met.

PR #3391, opened 2017-05-25
SUP-11057 ForceLayoutRedraw causes player not to load when hidden

ErezBouskila:


ErezBouskila:
Only do redraw if the player is visible in the DOM

PR #3396, opened 2017-05-29
Enable muted autoplay on mobile devices

dan-ziv:


modules/DoubleClick/resources/mw.DoubleClick.js
```diff
@@ -1159,7 +1163,11 @@
                 }
                 var size = _this.getPlayerSize();
                 _this.adsManager.resize( size.width, size.height, google.ima.ViewMode.NORMAL );
-                _this.adsManager.setVolume( _this.embedPlayer.getPlayerElementVolume() );
+                if ( _this.embedPlayer.mobileAutoPlay ) {
```
OrenMe: Why do we need to set it on loaded if on initialization we also do it?

dan-ziv: It's the original implementation, I just changed it accordingly.
Looks like a bug fix from the past: SUP-7515

PR #3403, opened 2017-05-30
2.56 dvr

eitanavgil:


modules/KalturaSupport/components/chapters/chapters.js
```diff
@@ -971,18 +981,35 @@
...
+		if ((this.getPlayer().isLive() && this.getPlayer().isDVR()) ||
+			!this.getPlayer().isLive()) {
 				//Send play request on first click for devices that don't have autoplay, e.g. mobile devices
 				if (!this.getPlayer().canAutoPlay() && this.getPlayer().firstPlay){
 					this.getPlayer().sendNotification('doPlay');
 				}
-				// see to start time and play ( +.1 to avoid highlight of prev chapter )
-				this.getPlayer().sendNotification('doSeek', ( this.mediaList[mediaIndex].startTime ) + 0.1);
+				if (this.embedPlayer.isDVR()) {
+					// get current time from Id3
```
kalturaguy: I think this code should be in a generic place so can be reused on other services
Like mw.id3tag.js .absolute2relative(...)

kalturaguy: Also you already have liveplayertime which contains the same calculation. So you can remove the code about id3 and yse it

modules/KalturaSupport/components/chapters/chapters.js
```diff
@@ -42,6 +42,7 @@
 
 		mediaList: [], //Hold the medialist items
 		chaptersMap: [],
+		currentId3Time: null,
```
kalturaguy: Plugin shouldnt know what  is id3 tag...

modules/KalturaSupport/components/chapters/chapters.js
kalturaguy: Also  id3 tag event is triggered when its parseed but you took current time do its 30-40 sec. Diff

modules/QnA/resources/qnaService.js
```diff
@@ -164,7 +164,10 @@
             return this.cuePoint().text;
         };
 
-        this.isAnswerOnAirActive = function(playerTime) {
+        this.isAnswerOnAirActive = function(playerTime,player) {
+            if(player && player.embedPlayer.isDVR()){
```
kalturaguy: This need to be fixed to use currentlivetime

PR #3410, opened 2017-06-04
Support dash manifest text/audio labels(FEC-6642)

itaykinnrot:
…on audio and text tracks - add it to out wrapper
Still pending shaka release  - tested locally with shaka master


OrenMe:
handled in https://github.com/kaltura/mwEmbed/pull/3412

PR #3424, opened 2017-06-15
with latest 2.58 (with DVR) - new polls API and polls push notifications

eitanavgil:


modules/KalturaSupport/resources/webcast/cuePointsManager.js
```diff
@@ -477,6 +516,7 @@
             return result;
         },
         getCurrentTime: function(){
+            //TODO - check if can replaced with isLive !!!
             if(this.embedPlayer.isDVR()){
```
eitanavgil: See if this can be replace with isLive()

PR #3435, opened 2017-06-22
Milticast webcast 2.58

eitanavgil:


modules/KalturaSupport/resources/mw.id3Tag.js
```diff
@@ -71,17 +71,22 @@
...
                 var d = new Date(0); // The 0 there is the key, which sets the date to the epoch
                 d.setUTCSeconds(time);
-                this.log("Update time from id3 tag: " + d.toUTCString());
+	            this.log("Update time from id3 tag: " + d.toUTCString());
                 this.updatedTime = time;
                 this.counter = 0; //reset time update interval counter
                 this.getPlayer().LiveCurrentTime = time;
+                if(this.getPlayer().isMulticast){
+	                //set the current time for MC from the ID3 tag and not from the SL element that has always 0
+	                this.getPlayer().currentTime = time
```
kalturaguy: this looks very odd....  places that uses currentTime should use LiveCurrentTime instead.

modules/KalturaSupport/resources/mw.id3Tag.js
```diff
@@ -71,17 +71,26 @@
         parseTag: function(tag){
             var time;
             if ( tag ) {
+                // on IE the tag needs to be parsed
                 time = tag.timestamp / 1000;
+                if(!time){
```
kalturaguy: This is wrong... 
Fixed the id3 tag and the correct place:
https://github.com/kaltura/mwEmbed/blob/db76a29d15909ba3f85fb65108f9ff302c88d900/modules/EmbedPlayer/resources/mw.EmbedPlayerSilverlight.js#L692

PR #3436, opened 2017-06-22
Fec 6621 - add ability to show/hide screen by external API 

eitanavgil:


modules/KalturaSupport/resources/mw.KBaseScreen.js
```diff
@@ -43,6 +43,16 @@
 			this.bind('playerReady', $.proxy(function (e, size) {
 				_this.error = false;
 			}, this));
+			this.bind('showScreenByName', $.proxy(function (e, screenName) {
+				if(this.pluginName == screenName){
+					this.showScreen();
```
OrenMe: Better add protection for both open and close like we have in the toggle method

OrenMe:
@eitanavgil always delete branch after PR is merged - this is for cleanup purposes

PR #3454, opened 2017-07-07
add a notification in related to notify other plugins that the relate…

eitanavgil:
…d-list is empty.


OrenMe:
@eitanavgil do you want to merge this already?

ErezBouskila:
@OrenMe @eitanavgil  Not sure if this is related but we added this improvement that if the list is empty or does not have related videos we disable the plugin.
https://github.com/kaltura/mwEmbed/pull/3388

eitanavgil:
@ErezBouskila  fo for it 👍  
Suggest a pull-request - I will be happy to review it

PR #3465, opened 2017-07-10
SUP-11360

eitanavgil:


modules/KalturaSupport/components/dualScreen/imagePlayer.js
```diff
@@ -104,6 +106,28 @@
...
+			this.bind('KalturaSupport_ThumbCuePointsReady' , function(){
+				if (_this.getPlayer().isLive()){
+					//Forcing 1st slide is only on VOD entry
+					return;
+				}
+				// on non-autoplayed player - check if required to force-show the 1st CP
+				var chaptersRawData = _this.getCuePoints();
+				//Since imagePlayer doesn't have a uiconf element - copy the value from dualScreen which has representation in the uiconf;
+				if (_this.embedPlayer.plugins.hasOwnProperty("chapters")
+					&& _this.embedPlayer.getKalturaConfig("chapters","showFirstSlideOnLoad")){
```
OrenMe: @eitanavgil why are you setting the config in chapters plugin, where it doesn't seem you use it, and then access him from the dualScreen image player in such a way?
Why not define it on dualScreen plugin and then pass it in imageAyer unit like we do with other config options?

eitanavgil: @OrenMe maybe this line 116 can explain it ;)
I do see a use-case where there is no dualScreen plugin and only chapters so I thought Chapters is better 
what do you think?

modules/KalturaSupport/components/dualScreen/imagePlayer.js
```diff
@@ -104,6 +106,28 @@
...
 			var _this = this;
 
+			this.bind('KalturaSupport_ThumbCuePointsReady' , function(){
+				if (_this.getPlayer().isLive()){
+					//Forcing 1st slide is only on VOD entry
+					return;
+				}
+				// on non-autoplayed player - check if required to force-show the 1st CP
+				var chaptersRawData = _this.getCuePoints();
+				//Since imagePlayer doesn't have a uiconf element - copy the value from chapters plugin which has representation in the uiconf;
```
OrenMe: @eitanavgil what do you mean it doesn't have representation? Anything under dualScreen is configured in the dualScreen plugin and passed to its child components when it initialize it

OrenMe: @eitanavgil when you init a class that extends from KBasePlugin, you pass to it a callback function as the second argument: https://github.com/kaltura/mwEmbed/blob/master/modules/KalturaSupport/resources/mw.KBasePlugin.js#L6
and then you can set it's default settings there, see example here(and also all over dualScreen plugin(that initializes imagePlayer component as well):
https://github.com/kaltura/mwEmbed/blob/2b0253b80c4ca0464c89cc9a5c3dc8ee61ca3265/modules/KalturaSupport/components/dualScreen/dualScreen.js#L488

So you can just add the desired config option under the dualScreen config, and when dualScreen initializes it's child component - imagePlayer - just pass the config there:
https://github.com/kaltura/mwEmbed/blob/2b0253b80c4ca0464c89cc9a5c3dc8ee61ca3265/modules/KalturaSupport/components/dualScreen/dualScreen.js#L876...L883
```javascript
var passthroughConfig = {
  prefetch: this.getConfig('prefetch'),
  cuePointType: this.getConfig('cuePointType'),
  showFirstSlideOnLoad: this.getConfig('showFirstSlideOnLoad')
};
```

eitanavgil:
@OrenMe  I applied your suggestion and now the parameter is in the dualScreen (Also added a Studio parameter. We changed the default behavior so now the Studio user can set it to not force 1st slide. Waiting for product to approve text.

OrenMe:
@eitanavgil hi, what is the status of this PR, can we merge?

OrenMe:
@eitanavgil please reopen when this PR is relevant again.

PR #3477, opened 2017-07-23
SUP-10955 Media preloads before pre roll with playlist

ErezBouskila:


modules/Hlsjs/resources/hlsjs.js
```diff
@@ -690,7 +690,15 @@
 				if (!this.mediaAttached){
 					this.unbind("firstPlay");
 					this.unbind("seeking");
-					this.hls.attachMedia(this.getPlayer().getPlayerElement());
+					if (this.embedPlayer.adTimeline) {
+						var self = this;
```
OrenMe: @ErezBouskila we can check for `firstPlay` as we handle it for other use cases.
Always register for it and then when we call play a few lines after the firstPlay event will be fired with regards of ad being player or not.

ErezBouskila:
@OrenMe made the change for firstPlay

PR #3487, opened 2017-07-30
SUP-11548 ForceLayoutRedraw playlist issue

ErezBouskila:
@OrenMe When using the flashvar with a playlist player the videoDisplay resizes to the entire playlistInterface width, if it is a playlist player I return it to the original size.


ErezBouskila:
@OrenMe Made the change and will let Dan test it out for us

PR #3503, opened 2017-08-10
Enable default captions from web sender

dan-ziv:


modules/Chromecast/resources/sender/mw.EmbedPlayerChromecast.js
```diff
@@ -571,8 +560,9 @@
         },
 
         getReceiverConfig: function () {
-            var chromecastConfig = this.getKalturaConfig( 'chromecast' );
-            return chromecastConfig.receiverConfig;
+            var receiverConfig = this.getKalturaConfig('chromecast').receiverConfig || {};
+            receiverConfig.defaultLanguageKey = this.beforeCastParams.captions[0] ? this.beforeCastParams.captions[0].lang : null;
```
OrenMe: Don't you need null/undefined protection? you did it before in `activeBeforeCastCaptions`

dan-ziv: done.
10x

PR #3507, opened 2017-08-15
Add iOS11 regex

dan-ziv:


OrenMe:
@dan-ziv why not add regex to detect the versions 

```javascript
var sOS=/OS ((\d+_?){2,3})\s/; //regex for OS string
var sVersion=/Version\/((\d+.?){2,3})\s/; //regex for Version string
var result = (sOS.exec(userAgent) || sVersion(userAgent)); //Match regex
var majorVersion;
if (result and result[1]){
   majorVersion = result[1].split(/[\._]/); //Split result of regex to get major version
}
```

With the above you can get the version number string and use it to check if result is >= 7

PR #3513, opened 2017-08-17
Rapt Media revisions and API features

khalsah:
#### What's this PR do?

- Expose API for Rapt Media Interactive Videos.
- Prevent a redundant initialization of the interactive video engine which was occurring.
- Parallelize initialization of segment data and engine initialization to improve load times.
- Fix declared dependencies of plugins which could lead to errors on initialization.
- Replay UI on end of project was removed because hitting the end of the playlist would incorrectly trigger the replay logic breaking the project. This should be re-added once a strategy for preventing media ended code from running at the end of the playlist is found.
- Logic was added to block the play action until the interactive video engine has been initialized.
- Overridden UI components were cleaned up to minimize duplicated code. They are still very tightly coupled with the underlying UI components. Ideally the underlying components would be extended to provide the necessary hooks so that we would not need to override them (which I expect will have maintenance problems at some point). That said my objective in this PR was to only touch code inside the rapt media module.
- Attempts to provide more comprehensive error messages, including checking for any missing segments at load time until waiting until the player attempts to use them.

#### Where should the reviewer start?

- `mw.RaptMedia.js`: as much of the interesting logic as possible has been moved into this file.
  - `KalturaSupport_EntryDataReady` is used to check for an interactive video project and enable the rapt engine if applicable.
  - `onChangeMedia` disables the rapt engine so that non-interactive videos can be played.
  - `prePlayAction` is used to block playback until the engine is ready.
  - Internally a config value `status` is maintained with the value of `disabled`, `loading`, or `enabled`.
  - A utility function `whenReady` will defer a function until the `raptMedia_ready` event if the engine is still loading.
  - `jQuery.Deferred` is used to provide a `Promise` like mechanism to make handling parallel loading of resources and the associated dependencies easier.
- `mw.RaptMediaPlayPauseBtn.js`: Not used in this version of the plugin, but required to maintain backwards compatibility with config's that are loading it.

#### What issues remain?

- [ ] The streamer type will sometimes not have been determined by the time we need it to compute the segment uncertainty at: https://github.com/FlixMaster/mwEmbed/blob/53970812530a722c999c43ac09159facd8322eae/modules/RaptMedia/resources/mw.RaptMedia.js#L355
**Is there an event we can wait for to ensure that the correct `streamerType` value will be available?**
- [ ] Testing with a heavily throttled CPU on Chrome will cause nodes incorrectly looping / auto replaying. **See more below** Update: This has been mitigated to only cause some stuttering at the end of nodes instead of a full loop.
- [ ] Safari sometimes refuses to seek to the requested time in an hls stream. We mitigate this by seeking to either the beginning or end of a segment if the current time is outside the current segment. With the changes to seeking behavior here safari seems more likely to get temporarily "stuck" seeking outside the current segment repeatedly, which makes node transitions on safari sometimes quite slow. Exact steps to reproduce are unknown, but it is more likely to occur when seeking to a part of the video that has not yet been buffered/loaded.
- [ ] During development situations have been observed where the video is playing, but the monitorEvent is not firing. Steps to reproduce are unknown.
**Is there any idea why the monitorEvent would not be firing?**
- [ ] hls.js logic for seeking over timeline "gaps" can conflict with our logic for seeking to the end of the segment if we're outside the bounds of the segment. If the end time of the segment is within one of those "gaps" hls.js and the rapt plugin will go into a continuous loop of seeking forward then backwards.
- [ ] Triggering a seek / `raptMedia_doJump` from the `raptMedia_ready` event will cause the player to completely lock up on mobile safari.

Some of these are likely pre-existing issues, and may not need to be fixed immediately.

#### Segments looping on slow machines

If the chrome performance dev-tools have performance throttled to a 20x slowdown, the end of a segment will cause issues including repeated stuttering, and eventually replaying the current node.

The follow code references seem to be contributing to this:

- https://github.com/FlixMaster/mwEmbed/blob/87fb173b5d22d62cbb1293dd3c2ca758476fa668/modules/RaptMedia/resources/mw.RaptMedia.js#L592
  When we detect the end of the segment is near, we seek to the exact end of the segment and request that the player "stop after seek".

- https://github.com/kaltura/mwEmbed/blob/3e6ffde5151238be4f1b5b0570379af394c95234/modules/EmbedPlayer/resources/mw.EmbedPlayer.js#L1178-L1182
  When the player believes the seek is finished, it pauses the player. _However_ it does this asynchronously. On slower machines it is possible that other code will have run in the mean time.

- https://github.com/kaltura/mwEmbed/blob/3e6ffde5151238be4f1b5b0570379af394c95234/modules/EmbedPlayer/resources/mw.EmbedPlayerNative.js#L1251
  Finally the code in the native player `_onpause` see's that a _new_ seek has been triggered, and therefor decides to trigger a `play` command. Removing it does appear to resolve the looping issue. **What bug or feature is this line needed for? Is there any way to bypass this logic?**


OrenMe:
Hi @khalsah. the PR is quite big, so I'll add notes as I see them.

For the issue of waiting for the rapt library to load before initiating the playback, e.g. the prePlayAction hook, you can just postpone the entire flow with the plugin async hook we have.
I think that if you use it you can also stop protecting all of the code with whenReady and additional protections I saw in your code cause you can be sure your SDK is loaded.
Just set `this.asyncInit = true;`in your plugin and when you are ready to continue call the `this.initCompleteCallback();` method and player will resume the flow.
What the above is doing to the player is:
1. player will load all plugins.
2. player will check when all plugins are initiated if they are also ready, in our case because you set asyncInit to true it will understand that it needs to wait.
3. once you call initCompleteCallback in your plugin it will single to player it is good to go.

Another thing I noticed, you can use `this.bind` and pass a space-delimited list of events.
It is just a convenience, but will save you some extra lines if all events call same handler.
This:
```javascript
this.bind('monitorEvent', function(){
	_this.updateEngine();
});
this.bind('onplay', function() {
	_this.updateEngine();
});

this.bind('onpause', function() {
	_this.updateEngine();
});
```

Can be written like this:
```javascript
this.bind('monitorEvent onplay onpause', function(){
	_this.updateEngine();
});
```

RaptMediaPlayPauseBtn -> if you deprecated it just remove it, the player won't crash of you add a setting that is not supported, it will just ignore it.

Custom event names: raptMedia_doPlay, raptMedia_doSeek, etc...
Are these exposed so you can call it via API from outside the player? haven't seen it being emitted from the code anywhere?

khalsah:
Thanks @OrenMe 

Both the plugin and our external engine have additional asynchronous requests they need to make on a _per entry_ basis not just to initialize their code. So, while I could use `asyncInit` to hold the initial load, it did not appear that I could use it to hold playback if the `changeMedia` command was used. Am I missing something there? If I do have to use `prePlayAction` for the `changeMedia` scenario is there any benefit to also adding the `asyncInit` code?

I made the tweaks to the `bind` calls, that's a lot cleaner.

I tried removing the unused plugin, however since it was used to override an existing plugin (`playPauseBtn`) existing player usage has the config `{ "playPauseBtn.plugin": false, "raptMediaPlayPauseBtn.plugin": true }`. So, removing the `raptMediaPlayPauseBtn` plugin leaves the player without any play/pause button.

Yes `raptMedia_doPlay`, etc. are indented for use outside the player.

OrenMe:
Hi @khalsah,

## async init and loading
I see you have two things loaded async:
* The rapt library - on very first load of the player you load it with `loadEngine`.
* The playlist segments - in any media loading, be it first media or change media flow, you call it in `loadSegments` and then call your engine, which seems async as well.

What you can do, which again, will save you some safety checks all around, but not necessarily save you code, will be to register to the `checkPlayerSourcesEvent` which is emitted in first load and in every changeMedia.
The event carries a callback argument which you need to call to signal the player that your data or setup is ready and player can proceed.
So if you add 
```javascript
embedPlayer.bindHelper( 'checkPlayerSourcesEvent', function( event, callback ) {
    _this.playbackCallback = callback;
});
```
then in your code you can do whatever you need and only when you call the `playbackCallback` you saved before will the player finish the loading, will it be first one or change media for consecutive plays. 

## The raptMediaPlayPauseBtn
I understand, and that is indeed a problem. If you have a limited amount of trials or haven't gotten to production yet then you can just edit the uiconfs that were already set like this and then you could remove this legacy support, otherwise if you are in production already then it will have to remain like this until you can come up with a deprecation path(notify clients so they can edit their uiconfs accordingly etc.)
I believe this can be managed with @zoharbabin.

## custom events
Feel free to extend as you require of course, just keep in mind that if you already have a player event that does the same you can use it. totally up to you.
I would be interested to know if you found an issue with current event scheme that made you extend the API.

khalsah:
@OrenMe 

## Async Initialization
Thanks for pointing out `checkPlayerSourcesEvent`, that helps clean things up. I've made those changes

## Deprecation
Getting rid of `raptMediaPlayPauseBtn` should be doable, we'll just have to do it in phases, starting by removing the functionality, then once that's live we can look at changing the configs, and then we can delete the file later.

## Custom Events
So, there are really three categories of events were adding:
1. New commands: `raptMedia_doCommand` and `raptMedia_doJump` these are for new functionality that only makes sense in the context of rapt media.
2. Commands we need to override: `raptMedia_doSeek` and `raptMedia_doReplay`. Seek is the hard one of these. In the context of an interactive video an API consumer would normally want to seek _within_ the current segment. There did not appear to be any way to intercept the original `doSeek` command to apply the additional logic to it, and even if we could I believe it would create headaches to differentiate between a seek in the playlist and a seek in the segment.
3. Commands for consistency: `raptMedia_doPlay` and `raptMedia_doPause`. It's true these are technically unnecessary as it stands today. However, it seemed like a better developer experience to have `raptMedia_` equivalents of these for consistency with the `raptMedia_doSeek` command.

OrenMe:
@khalsah glad the checkPlayerSourcesEvent was useful.
Regarding the seek issue, if you require manipulation of the seek behviour I suggest you pipe the seek of the player through you.
What you can do is: 
1. when you decide rapt should be initialised for you media grab a hold of the seek or doSeek method of the player, which one that you require(seems like seek in your case)
2. Override it with your own logic(the one you have in your plugin's seek method)
3. Call the original seek from your seek handler.
4. When media is changed and you decide rapt is not necessary anymore cleanup your override

I believe seek for the API level should always be a seek in the current segment, and you should only add new functionality to what is new in the player, which is the new playlist seek, which I wouldn't have a terminology of seek for it, but rather something in the form of clip change.
this is my two cents, but I believe this is more in your area of expertise J 

The only rational I can see is that if you leave the original API of the player as is, and it can do unexpected things then you better change it to do what you want cause users which will look for help in player docs/google/stackoverflow/etc will most likely see the original API docs.

Beside that LGTM, most of the other code is your own BL so I can't comment to much.

khalsah:
@OrenMe 

Okay, I've made the changes to override the `seek()` method to always seek within the current segment. I'm a touch nervous about that override interacting poorly with existing code, but it appears to be working.

I left the custom `raptMedia_doSeek` etc events in place for the time being since we have some references to them in various testing utilities on our end. Assuming problems with overriding the `seek` method don't crop up we can drop those commands in the future.

I also looked into overriding `getDuration()` since it currently returns the duration of the playlist, which doesn't make sense when `seek()` is overridden. However, I very quickly ran into some issues that didn't appear to be trivial to solve. I don't think trying to solve those issues makes sense for this PR given the urgency of getting it out.

I don't believe any of our current known issues are severe enough to block these changes from going in. We're running some QA now on the revisions to look for any new blocking problems. Are there any other blockers for merging this on your end?

OrenMe:
LGTM, give me the OK so I can merge.

calebhanson:
Oren, we are good to go from our end, you are good to merge this.

PR #3516, opened 2017-08-20
SUP-11400 Change dual screen and stream selector keyboardShortcutsMap

ErezBouskila:
@eitanavgil I talked with Karin about the keyboardShortcutsMap override and she said to remove the open close menu for the stream selector and to change the 'next state' and 'switch views' to 'z' and 'x'

### DIFF
--- a/modules/KalturaSupport/components/streamSelector.js
+++ b/modules/KalturaSupport/components/streamSelector.js
@@ -21,8 +21,6 @@
 				"nextStream": 221,   // Add ] Sign for next stream
 				"prevStream": 219,   // Add [ Sigh for previous stream
 				"defaultStream": 220, // Add \ Sigh for default stream
-				"openMenu": 83, // Add S Sigh for open menu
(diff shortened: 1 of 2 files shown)


modules/KalturaSupport/components/streamSelector.js
```diff
@@ -21,8 +21,6 @@
 				"nextStream": 221,   // Add ] Sign for next stream
 				"prevStream": 219,   // Add [ Sigh for previous stream
 				"defaultStream": 220, // Add \ Sigh for default stream
-				"openMenu": 83, // Add S Sigh for open menu
-				"closeMenu": "shift+83" // Add Shift+S Sigh for close menu
```
OrenMe: just remove them???

OrenMe: and does it make #3446 obsolete?

ErezBouskila: Yes we can remove them since the new interface of the stream selector (Now part of the dual screen new design) they do not preform any action. 

#3446 I closed.

PR #3534, opened 2017-09-10
FEC-7133 #comment extract frag from data object to avoid circular dep…

yairans:
…endency

### DIFF
--- a/modules/Hlsjs/resources/hlsjs.js
+++ b/modules/Hlsjs/resources/hlsjs.js
@@ -516,12 +516,26 @@
 					this.log("Try flash fallback");
 					this.fallbackToFlash();
 				} else {
-					var errorObj = {
-						message : JSON.stringify(data),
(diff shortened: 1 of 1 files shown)


modules/Hlsjs/resources/hlsjs.js
```diff
@@ -516,12 +516,26 @@
...
+					try {
+						var dataObj = {
+							type: data.type,
+							details: data.details,
+							fatal: data.fatal,
+							response: data.response,
+							networkDetails: data.networkDetails
+						};
+						var errorObj = {
+							message: JSON.stringify(dataObj),
```
OrenMe: add try catch for protection

yairans: is in try

OrenMe: :-( missed it

PR #3535, opened 2017-09-11
KMS-15253

MaksymKorniiov:
[IVQ] fix the responsive design for the question
https://kaltura.atlassian.net/browse/KMS-15253


OrenMe:
@eitanavgil - can we merge?

PR #3539, opened 2017-09-13
FEV-52

MaksymKorniiov:
https://kaltura.atlassian.net/browse/FEV-52 - add aria-valuemax, aria-valuetext and aria-valuemin

+ share screen:
1 - UI elements focused with tab order
2 - cycling focusing with TAB in share screen when it is opened
3 - when focusing with TAB on the URL lines - opened the times UI


OrenMe:
@MaksymKorniiov please add description to the PR so it is understood what the change is and why it is being done.

PR #3542, opened 2017-09-13
Scale images

jessp01:
See https://forum.kaltura.org/t/display-still-images-in-the-kaltura-player/5529/4


ErezBouskila:
@jessp01 With IE11 and Edge when going into fullscreen the image expands the entire width of the video element, I used a portrait image and it stretched out losing aspect ratio.

Object-fit is not supported in IE11 and Edge - http://caniuse.com/#search=object-fit

PR #3546, opened 2017-09-18
WEBC-529

MaksymKorniiov:
QNA Links should open in the user's default browser in a new tab. The link should be clickable.

### DIFF
--- a/modules/QnA/resources/qna.tmpl.html
+++ b/modules/QnA/resources/qna.tmpl.html
@@ -50,7 +50,8 @@
                         <div data-bind='css:{  "qnaAnnouncementText" : !isRead(),
                                                 "qnaAnnouncementTextRead" : isRead()
                                             },
-                                        text: entries()[0]().getText()'>
+                                        html: entries()[0]().getText(),
+                                        click: $parent.openLinkInNewTab'>
                         </div>
                         <br>
                         <div>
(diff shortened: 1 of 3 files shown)


modules/QnA/resources/qna.tmpl.html
```diff
@@ -50,7 +50,8 @@
                         <div data-bind='css:{  "qnaAnnouncementText" : !isRead(),
                                                 "qnaAnnouncementTextRead" : isRead()
                                             },
-                                        text: entries()[0]().getText()'>
+                                        html: entries()[0]().getText(),
+                                        click: $parent.openLinkInNewTab'>
```
eitanavgil: @MaksymKorniiov  
Does this resolve the UX of having pointer-cursor on the link?

modules/QnA/resources/qnaService.js
```diff
@@ -198,10 +198,15 @@
         };
 
         this.getText = function(){
+           var text = this.cuePoint().text+' ';
+            var reg = /(http:\/\/|https:\/\/).+?\s/gi;
```
eitanavgil: @MaksymKorniiov  
Please try to run different variations of URLs and see if they are all supported
Try http://www.something.com#abc and http://www.something.com?a=3 etc' and also check how this reacts with spaces.

PR #3569, opened 2017-10-08
Safari 11 autoplay fixes

dan-ziv:


OrenMe:
covered in separate commit to 2.62

PR #3571, opened 2017-10-10
locale fix

MaksymKorniiov:
if localization code of 5 char is set - grab only the first 2 
en_US will be en


modules/KalturaSupport/kalturaIframeClass.php
```diff
@@ -544,6 +544,7 @@ private function getLangKey(){
 		global $coreLanguageNames;
 		$playerConfig = $this->getUiConfResult()->getPlayerConfig();
 		if( isset( $playerConfig['vars']['localizationCode'] ) ){
+			$playerConfig['vars']['localizationCode'] =  strlen($playerConfig['vars']['localizationCode']) > 2 ? substr($playerConfig['vars']['localizationCode'], 0, 2) :  $playerConfig['vars']['localizationCode'];
```
OrenMe: @MaksymKorniiov extract method to method and wrote it in a more readable way.
Add method annotation to explain what purpose this method stands for

PR #3585, opened 2017-10-17
FEC-7288 FEC-7294 FEC-7295 FEC-7296 #comment handle ad sibling in saf…

yairans:
…ari 11


modules/KalturaSupport/resources/mw.KAdPlayer.js
```diff
@@ -1278,7 +1282,7 @@ mw.KAdPlayer.prototype = {
 			});
 			vid.src = source.getSrc();
 			vid.load();
-			vid.play();
+			vid.play().catch(doneCallback);
```
itaykinnrot: need to check if we have a promise here (IE8/9/10...)

yairans: done

PR #3588, opened 2017-10-17
KMS-15631 change to the Submit button

MaksymKorniiov:
[IVQ] change to the Submit button


modules/Quiz/resources/css/quiz.css
```diff
@@ -737,6 +737,13 @@ ol.second-row{
...
+    transition: width 0.3s,height 0.3s,top 0.3s,margin 0.3s ease-in-out;
+}
+.quizDone-cont.small{
+    width: 3vw;
+    height: 4vw;
+    top: 89vh;
+    margin-left: 97%;
```
Have we reviewed this on fullscreen? How does 97% look in fullscreen mode?

modules/Quiz/resources/quiz.js
```diff
@@ -737,6 +737,9 @@
                     .attr('title', 'click to end quiz now').on('keydown', _this.keyDownHandler).attr('id', 'quiz-done-continue-button').focus();
             // verify focus in IE
             document.getElementById('quiz-done-continue-button').focus();
+            setTimeout(function () {
+                $(_this.embedPlayer.getInterface()).find('.quizDone-cont').first().addClass('small');
+            },3000);
```
eitanavgil: @MaksymKorniiov  
If we are in playlist - do we need to clear this or is it destroyed naturally by the media change? (I am asking - not commenting)

OrenMe: @eitanavgil why don't you cache `quizDone-cont`?
you are creating it and then reffereing to it multiple times, but in each place you are using jquery selectors to reach it.
Save ref here:
```
scrubber.parent().prepend('<div class="quizDone-cont"></div>');
```
Then don't put global listeners here:
```
$(document).off( 'click', '.quizDone-cont' )
                 .on('click', '.quizDone-cont', function () {
                     _this.KIVQModule.quizEndScenario();
                 });
```
Unless I'm missing something, but why do we need to put it on window?
and then again in here:
```
$(_this.embedPlayer.getInterface()).find('.quizDone-cont').first().addClass('small');
```
You are searching for the same element again, where you just created it 10 lines above, so better just keep ref to it.

eitanavgil: @OrenMe  I agree. Although this function is called once. 

@MaksymKorniiov  can you please change that ?

PR #3591, opened 2017-10-18
FEV-119

MaksymKorniiov:
Quiz- Text is out of size both Russian & French

### DIFF
--- a/modules/Quiz/resources/css/quiz.css
+++ b/modules/Quiz/resources/css/quiz.css
@@ -372,6 +372,9 @@ button.qContinue, button.qApplied {
     vertical-align: middle;
     color: rgb(78, 166, 240);
 }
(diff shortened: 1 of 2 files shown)


modules/Quiz/resources/css/quiz.css
```diff
@@ -372,6 +372,9 @@ button.qContinue, button.qApplied {
     vertical-align: middle;
     color: rgb(78, 166, 240);
 }
+.hint-why-box.smaller-font {
+    font-size:1vw;
+}
```
As this ticket was opened on Russian and French - did we validate these specific languages ?

MaksymKorniiov: @eitanavgil I checked in quiz.js. If length of value longer than 4 character - I add this class.
In all of lang - text == 4 or lower. If it  bigger  - text will  be going outside of block

PR #3597, opened 2017-10-23
SUP-12276 Scrubber losing colour

ErezBouskila:
@OrenMe  Please review PR, the dualScreen css was overriding the playhead css so I just changed the css name.


OrenMe:
@eitanavgil the same UI class was meant so theming of player will also catch for other components, such as chapters.
Separating the UI classes might cause users who expect theming to apply to entire player to get a different result

OrenMe:
@eitanavgil @ErezBouskila what is the status of this, how do you handle the theming?

PR #3598, opened 2017-10-23
SUP-12305 Launch logo href using keyboard

ErezBouskila:


modules/KalturaSupport/components/logo.js
```diff
@@ -24,22 +24,22 @@
 							src: this.getConfig('img')
 						}).removeAttr("width height");
 				}
-				this.$el = $('<div />')
+				this.$el = $('<button />')
```
OrenMe: why change it to button? did you check the change on old IEs and mobiles?

ErezBouskila: @OrenMe Checked on All browsers and mobile devices and it works as expected, I changed it to a button because of this:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_button_role

PR #3599, opened 2017-10-25
add check if vid exists

eitanavgil:


OrenMe:
@eitanavgil the real issue here, I think, is that the method is even called on old browsers, as you are using flash so this calss shouldn't even be loaded.
I would go an check that.

PR #3601, opened 2017-10-27
Accessibility

MaksymKorniiov:


OrenMe:
@MaksymKorniiov please avoid naming PR with version name, nothing guarantees this will be a part of any version so this can be misleading.
When naming a PR please make it as informative and short as possible in the title, and add description in the body(first comment) describing the change

modules/KalturaSupport/components/volumeControl.js
```diff
@@ -158,6 +158,7 @@ mw.PluginManager.add( 'volumeControl', mw.KBaseComponent.extend({
 			var title = gM('mwe-embedplayer-volume-value', percent );
             var $slider = this.getSlider().find('a');
             this.setAccessibility($slider, title);
+            this.updateVolumeUI(percent/100);
```
eitanavgil: @MaksymKorniiov  
Did we check if the volume comes from the user cookie it is being updated accordingly?

modules/Quiz/resources/css/quiz.css
```diff
@@ -231,7 +233,7 @@
     cursor: default!important;
 }
 .display-question:focus{
-    outline: none;
+    /*outline: none;*/
```
If we decide that this is not relevant anymore - let's delete it.

modules/KalturaSupport/components/volumeControl.js
MaksymKorniiov: @eitanavgil Yes

modules/Quiz/resources/css/quiz.css
MaksymKorniiov: @eitanavgil fixed it in new commit

modules/KalturaSupport/components/logo.js
```diff
@@ -9,9 +9,9 @@
 			displayImportance: 'low',
 			align: "right",
 			cssClass: "kaltura-logo",
-			href: null,
-			title: null,
-			img: null
+			href: 'http://www.kaltura.com',
+			title: 'Kaltura',
+			img: 'http://kgit.html5video.org/branches/master//skins/kdark/images/kalturaLogo.png'
```
eitanavgil: @MaksymKorniiov  
We cannot hard-code a release with path to kgit. Let's try to change this to a relative path and if not possible I will work with you on an absolute kaltura CDN file.

PR #3606, opened 2017-11-02
fix(SUP-12496): ad noticeMessage appears only for preroll

ErezBouskila:
@OrenMe  Please review PR, The noticeMessage only appears for preroll and now it will show for all ads. I confirmed on my end that it works with pre mid and post.

### DIFF
--- a/modules/DoubleClick/resources/mw.DoubleClick.js
+++ b/modules/DoubleClick/resources/mw.DoubleClick.js
@@ -1253,6 +1253,7 @@
                 if ( _this.isLinear ) {
                     if ( !_this.adSkippable ) {
                         _this.showSkipBtn();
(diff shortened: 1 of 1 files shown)


dan-ziv:
@ErezBouskila 
We are already doing it in `getAdContainer`
Why this is not enough?

ErezBouskila:
@OrenMe @dan-ziv getAdContainer is only called one time, that is way the countdown notice is only showed for the preroll.

dan-ziv:
@ErezBouskila 
So we can remove the call from `getAdContainer`?
Seems its redundant now.

ErezBouskila:
@dan-ziv I will remove it and check

ErezBouskila:
@dan-ziv Removed and it works.

PR #3608, opened 2017-11-04
feat: only throw critical Shaka errors

OrenMe:
throwing recovarable errors causes player UI to show error message although Shaka can still recover


itaykinnrot:
@OrenMe maybe add a log for the non critical errors?

OrenMe:
closes FEC-7412

PR #3634, opened 2017-12-04
FEC-7267 #comment play once youtube promise is resolved

yairans:


modules/ExternalPlayers/resources/mw.EmbedPlayerYouTube.js
```diff
@@ -167,6 +167,7 @@
 				}
 				//autoplay
 				if(mw.getConfig('autoPlay') && _this.canAutoPlay()){
+					_this.unbindHelper("preSeek");
```
OrenMe: why is this required?

yairans: in case the `onAdError` triggered early this line -  https://github.com/kaltura/mwEmbed/blob/77d1d90c370813dd553f11a4f5f3ce53ce1c8a87/modules/AdSupport/resources/mw.AdTimeline.js#L463 before this - https://github.com/kaltura/mwEmbed/blob/77d1d90c370813dd553f11a4f5f3ce53ce1c8a87/modules/AdSupport/resources/mw.AdTimeline.js#L167
then, the first seek on playback doesn't work.
I tested this change on multiple cases: ad success, ad error and no ads.

PR #3638, opened 2017-12-05
fix(FEC-7556): handle restore player on ad error of LOG type

dan-ziv:


modules/DoubleClick/resources/mw.DoubleClick.js
```diff
@@ -1321,8 +1322,20 @@
                 mw.log( "DoubleClick:: adSkipped" );
                 $( _this.embedPlayer ).trigger( 'onAdSkip' );
             } );
+
+            adsListener('LOG', function (event) {
+                if (_this.nonFatalError) return;
```
OrenMe: print it anyway, so we can see in log

modules/DoubleClick/resources/mw.DoubleClick.js
```diff
@@ -1697,6 +1710,16 @@
         },
         // Handler for various ad errors.
         onAdError: function( errorEvent ) {
+            if (this.nonFatalError) {
```
OrenMe: Maybe better to create a `onNonFatalAdError` method, as the handling is quite unique for fatal and non fatal errors.

modules/DoubleClick/resources/mw.DoubleClick.js
```diff
@@ -1321,8 +1322,19 @@
                 mw.log( "DoubleClick:: adSkipped" );
                 $( _this.embedPlayer ).trigger( 'onAdSkip' );
             } );
+
+            adsListener('LOG', function (event) {
+                if (_this.nonFatalError) return;
+                var adData = event.getAdData();
+                if (adData['adError']) {
+                    console.log('Non-fatal error occurred: ' + adData['adError'].getMessage());
```
OrenMe: @dan-ziv print this anyway and just call `handleNonFatalError` conditionally.

PR #3644, opened 2017-12-06
FEV-141 add alert to banSeek feature 

MaksymKorniiov:
[IVQ-Player] no seek - alert

### DIFF
--- a/modules/EmbedPlayer/resources/mw.PlayerLayoutBuilder.js
+++ b/modules/EmbedPlayer/resources/mw.PlayerLayoutBuilder.js
@@ -1437,16 +1437,35 @@
          *
          * @param (Object) Object which includes:
          *   text message in popup
-         *   timeout time in seconds to hide popup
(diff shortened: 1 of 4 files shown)


modules/EmbedPlayer/resources/mw.PlayerLayoutBuilder.js
```diff
@@ -1432,6 +1432,27 @@
...
+         *   timeout time in seconds to hide popup
+         */
+        showMessage: function (data) {
+            if(data.text && data.timeout){
+                var $text = $('<p/>').addClass('text').text(data.text);
+                var $container = $('<div />').addClass('soft-message').append($text);
+                $(this.embedPlayer).append($container);
+                setTimeout(function () {
+                    $container.fadeOut("hide");
+                }, data.timeout * 1000)
```
eitanavgil: @MaksymKorniiov 
If there is no value in data.timeout, what will happen?

PR #3647, opened 2017-12-07
dualScreen-icons-accessible

MaksymKorniiov:
add focus by tab to 3 icons on dualScreen


modules/KalturaSupport/components/dualScreen/dualScreen.css
```diff
@@ -155,7 +155,7 @@
     visibility: visible;
     opacity: 1;
 }
-.componentOff{
+.componentOff *{
```
eitanavgil: @MaksymKorniiov 
I don't like using * - it is too risky. Any chance to be more protective of the entire CSS scope?

PR #3659, opened 2017-12-21
WEBC-1059: Slide broadcast twice if starting streaming with slide when two producers are opened

MaksymKorniiov:
v3.0.71| Slide broadcast twice if starting streaming with slide when
two producers are opened


modules/KalturaSupport/resources/mw.KCuePoints.js
```diff
@@ -405,6 +409,36 @@
...
+				if(isTheSamePartnerData && isTheSameTags && (differentByCreatedAt <= 3000 || differentByStartTime <= 3000)){
+					return false;
+				}
+			}
+			return true;
+		},
+		getPrevCPWithCorrectType: function (allCP,currentCuePointIndex) {
+			var prevCP = false;
+			for(var i = currentCuePointIndex; i>=0;i--){
+				if(allCP[i].cuePointType === "thumbCuePoint.Thumb"){
```
eitanavgil: @MaksymKorniiov 
why not if(allCP[i].cuePointType === currentCuePointIndex.cuePointType) ?

modules/KalturaSupport/resources/mw.KCuePoints.js
```diff
@@ -405,6 +409,36 @@
...
+		 *
+		 */
+		validateSameCuePoints:function (allCP, currentCuePointIndex) {
+			var currentCP = allCP[currentCuePointIndex];
+			var prevCP = this.getPrevCPWithCorrectType(allCP,currentCuePointIndex-1);
+			if(prevCP !== false && currentCP){
+				var differentByCreatedAt = Math.abs(currentCP.createdAt - prevCP.createdAt);
+				var differentByStartTime = Math.abs(currentCP.startTime - prevCP.startTime);
+				var isTheSamePartnerData = currentCP.partnerData === prevCP.partnerData;
+				var isTheSameTags = currentCP.tags === prevCP.tags;
```
eitanavgil: @MaksymKorniiov if we remove this line it will also cover admin-cp.
when an admin-user switches to live he sends admin-tagged-start CP and immediately public CP and therefore admin-users sees 2 slides. Removing this should fix also admin duplicity

modules/KalturaSupport/resources/mw.KCuePoints.js
```diff
@@ -405,6 +409,36 @@
...
+		 * if have same CP earlier - hide current CuePoint
+		 * @param  allCP - array of CP where try to find same CP
+		 * @param  currentCuePointIndex - position from current CP
+		 *
+		 */
+		validateSameCuePoints:function (allCP, currentCuePointIndex) {
+			var currentCP = allCP[currentCuePointIndex];
+			var prevCP = this.getPrevCPWithCorrectType(allCP,currentCuePointIndex-1);
+			if(prevCP !== false && currentCP){
+				var differentByCreatedAt = Math.abs(currentCP.createdAt - prevCP.createdAt);
```
eitanavgil: We usually use the phrase 'delta' to mark something that is the difference between 2 values E.G. createdAtDelta

modules/KalturaSupport/resources/mw.KCuePoints.js
eitanavgil: @MaksymKorniiov I see that I told you via the ticket to use cue-point with type thumb 
Let's apply the comment above anyway :)

modules/KalturaSupport/components/chapters/chapters.js
```diff
@@ -462,8 +462,12 @@
 
 			//Get current item number
 			var orderId = this.mediaList.length;
+			
+			var clearSameCP = items.filter(function( item,index,allInArray ) {
```
eitanavgil: @kalturaguy 
Do you think we should optimize this? From how it is it looks like we run this thing with EVERY CP that is processed in the logic

modules/KalturaSupport/resources/mw.KCuePoints.js
```diff
@@ -405,6 +409,36 @@
...
+				var differentByStartTime = Math.abs(currentCP.startTime - prevCP.startTime);
+				var isTheSamePartnerData = currentCP.partnerData === prevCP.partnerData;
+				var isTheSameTags = currentCP.tags === prevCP.tags;
+				if(isTheSamePartnerData && isTheSameTags && (differentByCreatedAt <= 3000 || differentByStartTime <= 3000)){
```
eitanavgil: @MaksymKorniiov  
What is the effort of getting the #3000 from configuration? 
It will give us the ability to fine-tune the feature without releasing new version. The default should be 3000

modules/KalturaSupport/resources/mw.KCuePoints.js
```diff
@@ -405,6 +409,54 @@
...
+				}
+			}
+			return true;
+		},
+		getPrevCPWithCorrectType: function (allCP,currentCuePointIndex) {
+			var prevCP = false;
+			var previousIndex = currentCuePointIndex - 1;
+			var currentCP = allCP[currentCuePointIndex];
+			for(var i = previousIndex; i>=0;i--){
+				var defaultThreshold = this.getThreshold();
```
eitanavgil: @MaksymKorniiov 
this.getThreshold() is called every iteration, but it never changes. If you have something that you can calculate before the iteration do it outside. With this implementation we call an unnecessary function every iteration.

eitanavgil: Also, since the Threshold is pretty static and will never change @ runtime - we should calculate this once @ load time and not recalculate it

modules/KalturaSupport/resources/mw.KCuePoints.js
```diff
@@ -405,6 +409,54 @@
...
 			return foundAttr;
 		},
+		/**
+		 * if have same CP earlier - hide current CuePoint
+		 * @param  allCP - array of CP where try to find same CP
+		 * @param  currentCuePointIndex - position from current CP
+		 *
+		 */
+		removeDuplicatedCuePoints:function (allCP, currentCuePointIndex) {
+			var defaultThreshold = this.getThreshold();
```
eitanavgil: @MaksymKorniiov  do this @ load time. This never changes

modules/KalturaSupport/resources/mw.KCuePoints.js
```diff
@@ -405,6 +409,54 @@
...
+			}
+			return true;
+		},
+		getPrevCPWithCorrectType: function (allCP,currentCuePointIndex) {
+			var prevCP = false;
+			var previousIndex = currentCuePointIndex - 1;
+			var currentCP = allCP[currentCuePointIndex];
+			for(var i = previousIndex; i>=0;i--){
+				var defaultThreshold = this.getThreshold();
+				var createdAtDelta = Math.abs(currentCP.createdAt - allCP[i].createdAt);
```
kalturaguy: it's wrong to to look at createdAt  in VOD they can be the same
check only startTime

modules/KalturaSupport/components/chapters/chapters.js
eitanavgil: @guy1976 we checked and this happens on CP load and only once, so we are good here

modules/KalturaSupport/resources/mw.KCuePoints.js
eitanavgil: We are calling this.getThreshold() with each iteration. If we calculate this @ load time only once it will be more efficient

PR #3661, opened 2017-12-25
SUP-12806-Video not loading automatically on iPad and iPhone after bu…

ErezBouskila:
…mper using Safari on iOS11


ErezBouskila:
@OrenMe Works with WebKitPlaysInline and changed to isIOS

PR #3688, opened 2018-01-15
add quality watcher for dash streams

itaykinnrot:


modules/KalturaSupport/resources/mw.QualityWatcher.js
```diff
@@ -0,0 +1,106 @@
...
+            var activeTrackId = 0;
+            for (var index=0 ; index < sortedTracks.length ; index++){
+                if (sortedTracks[index].active){
+                    activeTrackId = index;
+                }
+            }
+            return activeTrackId;
+        },
+        removeQuality:function(level){
+            var sortedTracks = this.getSortedTracks();
```
OrenMe: maybe need to check that we don't remove the latest available track, or at least handle the error state if a machine reaches a state where video playback is just not feasible and throw an error

modules/KalturaSupport/resources/mw.QualityWatcher.js
```diff
@@ -0,0 +1,106 @@
...
+            var bandWidthToRemove = sortedTracks[level].bandwidth;
```
OrenMe: I think you can get the current variant bandwidth can be achived also by calling `getStats` API.
The value is under `streamBandwidth`, so you might be able to avoid all the logic to look for current video track.

modules/KalturaSupport/resources/mw.QualityWatcher.js
itaykinnrot: taking care of that

PR #3689, opened 2018-01-15
fix(SUP-13171): adCover blocking VPAID controls

ErezBouskila:
@OrenMe Please review PR

### DIFF
--- a/modules/DoubleClick/resources/mw.DoubleClick.js
+++ b/modules/DoubleClick/resources/mw.DoubleClick.js
@@ -1326,6 +1326,9 @@
             // Resume content:
             adsListener( 'SKIPPED', function () {
                 mw.log( "DoubleClick:: adSkipped" );
+                if(_this.isVPAID === true) {
+                    _this.isVPAID = false;
+                }
                 $( _this.embedPlayer ).trigger( 'onAdSkip' );
             } );
 
@@ -1341,6 +1344,9 @@
             // Resume content:
             adsListener( 'CONTENT_RESUME_REQUESTED', function () {
                 if (_this.nonFatalError) return;
+                if(_this.isVPAID === true) {
+                    _this.isVPAID = false;
+                }
                 $( _this.embedPlayer ).trigger( 'onContentResumeRequested' );
                 _this.playingLinearAd = false;
                 // Update slot type, if a preroll switch to midroll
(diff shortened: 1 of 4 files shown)


ErezBouskila:
@OrenMe I can use the getContentType() and check the string of the ad.

OrenMe:
@ErezBouskila if this uses a valid API and official response then it’s fine

ErezBouskila:
@OrenMe Made the change, here is the list of the getContentType() types:
https://groups.google.com/forum/#!topic/ima-sdk/TaKullfdqH4

Should we check for application/x-shockwave-flash?

OrenMe:
@ErezBouskila I don't think this is safe either - unfortunately ad space is very fragmented and you can't be sure that ad networks or vendors return correct metadata.

ErezBouskila:
@OrenMe It seems that ad.getContentType() is the only way to use the API to know if the add is vpaid:
https://groups.google.com/forum/#!topic/ima-sdk/bTy9l2CSHOs

ErezBouskila:
@OrenMe Please review the latest fix.
Thanks

OrenMe:
@yairans please review as well

modules/DoubleClick/resources/mw.DoubleClick.js
```diff
@@ -1556,6 +1578,16 @@
                 'height': this.embedPlayer.getVideoHolder().height()
             };
         },
+        hideControlsOnVPAID: function(){
+            var hideControls = "-40px";
```
OrenMe: what happens if we are with actual hover controls?
also, `-40px` - what if we have custom CSS?

OrenMe: Maybe consider moving this to controlBars classes as methods and doing `this.embedPlayer.layoutManager.hideControls()` kind of a thing

ErezBouskila: @OrenMe I checked with hover controls and it works, I can make the pixel size larger as I do not want to mess around with z-index.

Will move it to controlBars

itaykinnrot:
Lets add a flag here just to be on the safe side

ErezBouskila:
@itaykinnrot @yairans Please review PR.
Thanks

ErezBouskila:
@OrenMe  @yairans  Made the final change (Hopefully), please review PR
Thanks

PR #3696, opened 2018-01-22
fix(SUP-13105): Rebuild CC menu on change media with external CC

ErezBouskila:
@OrenMe Please review PR, adding the destroy to the onChangeMedia did not work.

### DIFF
--- a/modules/KalturaSupport/resources/mw.ClosedCaptions.js
+++ b/modules/KalturaSupport/resources/mw.ClosedCaptions.js
@@ -83,6 +83,12 @@
 			}
 
 			if ( this.getConfig('showEmbeddedCaptions') === true ) {
+                this.bind( 'playerReady', function () {
+                    _this.destory();
+                    _this.setupTextSources( function () {
+                        _this.buildMenu( _this.textSources );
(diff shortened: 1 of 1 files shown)


modules/KalturaSupport/resources/mw.ClosedCaptions.js
```diff
@@ -360,6 +360,13 @@
 			}
 			this._super( property, value );
 		},
+		updateCaptionsMenu: function () {
+			var _this = this;
```
OrenMe: @ErezBouskila this is not required for embedded captions so why add it?

OrenMe: In 'showEmbeddedCaptions' the trigger to creating and populating the menu is from the event `textTracksReceived`
With this change we are new creating an empty menu on playerReady when we have embedded captions.

OrenMe: @ErezBouskila you need to just destroy the menu on changeMedia
Rebuilding it is handled separately for each mode and you need to keep it like that!

PR #3703, opened 2018-01-23
supporting banSeekManager plugin from Editor

MaksymKorniiov:
use banSeekManager if add status for it in Editor

### DIFF
--- a/modules/Quiz/resources/IVQModule.js
+++ b/modules/Quiz/resources/IVQModule.js
@@ -53,11 +53,28 @@
                     }
                     else {
                         $.quizParams = data[1];
+                        var banSeekTrigger = false;
+                        var dataForBanSeek = {
+                            status:false,
+                            alertText:''
+                        };
                         $.grep($.quizParams.uiAttributes, function (e) {
+                            
+                            if (e.key == "banSeek" && e.value) {
+                                banSeekTrigger = true;
+                                dataForBanSeek.status = true;
+                            }
+                            if (e.key == "alertText") {
+                                dataForBanSeek.alertText =  e.value;
+                            }
                             if (e.key == "canSkip") {
                                 _this.canSkip = (e.value.toLowerCase() === 'true');
                             }
                         });
+                        //send notification to banSeekManager with params from Editor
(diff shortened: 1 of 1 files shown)


modules/Quiz/resources/IVQModule.js
```diff
@@ -53,11 +53,28 @@
                     }
                     else {
                         $.quizParams = data[1];
+                        var banSeekTrigger = false;
+                        var dataForBanSeek = {
```
eitanavgil: @MaksymKorniiov  
To me it looks that there is a redundant variable here. Do we need banSeekTrigger if we are also saving the same value in dataForBanSeek?

modules/Quiz/resources/IVQModule.js
```diff
@@ -53,11 +53,28 @@
...
+                            if (e.key == "alertText") {
+                                dataForBanSeek.alertText =  e.value;
+                            }
                             if (e.key == "canSkip") {
                                 _this.canSkip = (e.value.toLowerCase() === 'true');
                             }
                         });
+                        //send notification to banSeekManager with params from Editor
+                        if(banSeekTrigger && !_this.canSkip){
+                            _this.embedPlayer.sendNotification('activateBanSeek',dataForBanSeek);
```
eitanavgil: @OrenMe  
what do you think of this approach? We are triggering a notification that will react only on an external plugin served from kwidgetPs. If it is missing from player nothing will happen.

modules/Quiz/resources/IVQModule.js
```diff
@@ -53,11 +53,28 @@
...
+                        var banSeekTrigger = false;
+                        var dataForBanSeek = {
+                            status:false,
+                            alertText:''
+                        };
                         $.grep($.quizParams.uiAttributes, function (e) {
+                            
+                            if (e.key == "banSeek" && e.value) {
+                                banSeekTrigger = true;
+                                dataForBanSeek.status = true;
```
eitanavgil: @MaksymKorniiov  - I see that a few lines below there is a e.value.toLowerCase() 
Do you see a reason for use/not use the same protection here?

PR #3706, opened 2018-01-24
change the returned data from kava and add flag for monitorViewEvents

itaykinnrot:


itaykinnrot:
Dont merge till @erankor  approve

erankor:
@itaykinnrot, a couple of comments -
1. The code has to support both formats for now, since we can't really make the player+server change together (due to caching). So we need something like - `if (typeof parsedData === 'object') { new code } else { old code }`. 
2. I want the default to be 'enabled', currently if the property is not sent by the server it will be disabled. So either - a. you'll check whether the property exists or b. we'll change the definition to viewEventsDisabled. Let me know what you prefer.
3. The 3 !'s looks a bit redundant :) one would suffice.

itaykinnrot:
fixed @erankor

erankor:
Looks good to me, you can probably simplify the condition to viewEventsEnabled === false, but not really critical...

PR #3713, opened 2018-01-25
fix: do not automute for autoplay on chrome64

OrenMe:
the autoplay restriction has been defered to version 66 of chrome


modules/MwEmbedSupport/mediawiki/mediawiki.client.js
```diff
@@ -117,10 +117,10 @@
 	mw.isChrome = function () {
 		return ( userAgent.indexOf('Chrome') != -1 && !mw.isEdge() );
 	};
-    mw.isChrome64AndUp = function () {
+    mw.isChromeVersionGreaterThen = function (version) {
```
yairans: should `isChromeVersionGreaterThan` with `a`

PR #3737, opened 2018-02-18
fix(SUP-13280): Mobile skin is blocking the YouTube player play button

ErezBouskila:
@OrenMe  Please review PR.


OrenMe:
@ErezBouskila I see no relation between title and fix and original bug

PR #3754, opened 2018-03-11
operator is not valid for PHP string concatenation

DBezemer:
Causes immediate invalidation of every entry object in cache for every request.
Validated both with on disk and memcache cache backend


DBezemer:
@OrenMe can you please merge? This currently breaks the player when used on PHP 7.1

OrenMe:
@DBezemer thanks! will be added in 2.68

PR #3759, opened 2018-03-14
fix(SUP-13714): Disable below layout with embedded CC

ErezBouskila:
@OrenMe Please review for 2.68, There is no need for layout below with embedded CC as they do not use the allocated space and they are added to the shadow DOM.


ErezBouskila:
@OrenMe Can you please review this PR.
Thanks

PR #3761, opened 2018-03-14
Fev 150 block portrait mode

MaksymKorniiov:


modules/Quiz/resources/quiz.js
```diff
@@ -29,6 +29,7 @@
         multiStreamWelcomeSkip:false,
         relatedStreamChanging:false,
         IVQVer:'IVQ-2.41.rc2',
+        ivqShowScreenMode: false,
```
eitanavgil: ShowScreen is too generic. This is a vertical-block screen. Let's try 'mobileWarningLayer' or something that will imply the meaning of the screen

modules/Quiz/resources/quiz.js
```diff
@@ -218,6 +219,13 @@
                     };
                 }
             });
+            $( window ).on( "orientationchange", function() {
```
eitanavgil: Not sure - but maybe we should add this event-listening only on mobile?

modules/Quiz/resources/quiz.js
```diff
@@ -218,6 +219,13 @@
...
+                if(_this.ivqShowScreenMode && _this.canShowPortraitWarning() ) {
```
eitanavgil: canShowPortraitWarning to isPortrait ? 
also, maybe there is an existing check today on the CSS class level? (not sure - asking)

PR #3762, opened 2018-03-18
feat: kava addons

dan-ziv:


modules/KalturaSupport/resources/mw.kAnalony.js
```diff
@@ -100,14 +105,17 @@
...
+                _this.resetPlayerflags();
 				if ( _this.kalturaContextData && _this.kalturaContextData.flavorAssets && _this.kalturaContextData.flavorAssets.length === 1 ){
 			        _this.currentBitRate = _this.kalturaContextData.flavorAssets[0].bitrate;
 		        }
 				_this.sendAnalytics(playerEvent.IMPRESSION);
 			});
 
 			this.embedPlayer.bindHelper( 'onChangeMedia' , function () {
+				_this.timer.destroy();
+				_this.rateHandler.destroy();
```
OrenMe: Usually you would expect an instance to not be useable after calling destroy on it.
If you mean reset then rename it to that.
Destroy should be used for cleanup.

PR #3767, opened 2018-03-21
[RaptMedia] Support pure-Kaltura projects and vendor engine

khalsah:
- Implements support for the pure Kaltura version of Rapt Media interactive video, by checking for the magic value `raptmedia!` in the `partnerData` and loading the graph data via the `fileAsset` service. **Note** This depends on a change to the Kaltura API which is currently only deployed to a staging server.
- Vendors the interactive video engine inside the plugin.


modules/RaptMedia/resources/mw.RaptMedia.js
```diff
@@ -157,8 +156,12 @@
 		},
 
 		readRaptProjectId: function() {
-			var partnerData = this.getPlayer().evaluate('{mediaProxy.entry.partnerData}');
-			var segments = (partnerData || "").split(';');
+			var partnerData = this.getPlayer().evaluate('{mediaProxy.entry.partnerData}') || '';
+			if (partnerData.toLowerCase() === 'raptmedia!') {
```
gonenradai: is this used for some kind of distinction between old and new playlists?

khalsah: Yes. For existing (old) rapt backend based projects this field contains `raptmedia;[RAPT PROJECT ID]`.

This adds a check for `raptmedia!` which it treats as a (new) kaltura based project.

Is there a better field to use for this?

PR #3772, opened 2018-03-22
fix(FEC-8072): KAVA - "eventIndex" and "playTimeSum" parameters aren't reset after change media

dan-ziv:


modules/KalturaSupport/resources/mw.kAnalony.js
```diff
@@ -115,6 +115,8 @@
 			this.embedPlayer.bindHelper( 'onChangeMedia' , function () {
 				_this.timer.destroy();
 				_this.rateHandler.destroy();
+                _this.eventIndex = 1;
+                _this.playTimeSum = 0;
 				_this.bufferTimeSum = 0;
```
itaykinnrot: @dan-ziv should we also include bufferTime?

PR #3778, opened 2018-03-28
KMS-17301-true-false-inline-position:

MaksymKorniiov:
fix margin between true/false answers

### DIFF
--- a/modules/Quiz/resources/css/quiz.css
+++ b/modules/Quiz/resources/css/quiz.css
@@ -865,7 +865,7 @@ ol.second-row{
 .true-false-question .single-answer-box-bk{
     width: 41vw;
     float: left;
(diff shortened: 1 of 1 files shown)


modules/Quiz/resources/css/quiz.css
```diff
@@ -865,7 +865,7 @@ ol.second-row{
 .true-false-question .single-answer-box-bk{
     width: 41vw;
     float: left;
-    margin: 0 22px;
+    margin: 0 3vw;
```
eitanavgil: @MaksymKorniiov  - can we also verify this on fullscreen?

MaksymKorniiov: @eitanavgil it works and looks good

PR #3786, opened 2018-04-11
FEV-175-fix-filtering:

MaksymKorniiov:
1. not filtering if partnerData empty
2. add checking for  same title and same description
3. remove createdAtDelta checking

### DIFF
--- a/modules/KalturaSupport/resources/mw.KCuePoints.js
+++ b/modules/KalturaSupport/resources/mw.KCuePoints.js
@@ -457,12 +457,13 @@
 			var defaultThreshold = this.threshold;
 			var currentCP = allCP[currentCuePointIndex];
 			var prevCP = this.getPrevCPWithCorrectType(allCP,currentCuePointIndex);
-			if(prevCP !== false && currentCP){
-				var createdAtDelta = Math.abs(currentCP.createdAt - prevCP.createdAt);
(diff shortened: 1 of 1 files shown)


modules/KalturaSupport/resources/mw.KCuePoints.js
```diff
@@ -457,12 +457,13 @@
...
-			if(prevCP !== false && currentCP){
-				var createdAtDelta = Math.abs(currentCP.createdAt - prevCP.createdAt);
+			if(prevCP !== false && currentCP && currentCP.partnerData){
 				var startTimeDelta = Math.abs(currentCP.startTime - prevCP.startTime);
 				var isTheSamePartnerData = currentCP.partnerData === prevCP.partnerData;
+				var isSameTitle = currentCP.title === prevCP.title;
+				var isSameDescription = currentCP.description === prevCP.description;
 				var isTheSameTags = currentCP.tags === prevCP.tags;
-				if(isTheSamePartnerData && isTheSameTags && (createdAtDelta <= defaultThreshold*1000 || startTimeDelta <= defaultThreshold*1000)){
+				if(isTheSamePartnerData && isTheSameTags && isSameTitle && isSameDescription && startTimeDelta <= defaultThreshold*1000){
```
kalturaguy: @eitanavgil  I think the tags will be different in webcsat, since one slide could be in preview and the other not

PR #3791, opened 2018-04-15
add 1 min cache

eitanavgil:


modules/KalturaSupport/resources/mw.KCuePoints.js
```diff
@@ -228,7 +228,9 @@
 			var lastCreationTime = _this.getLastCreationTime() + 1;
 			// Only add lastUpdatedAt filter if any cue points already received
 			if (lastCreationTime > 0) {
-				request['filter:createdAtGreaterThanOrEqual'] = lastCreationTime;
+				var cpThreshold  = mw.getConfig("cuePointsThreshold") ? parseInt(mw.getConfig("cuePointsThreshold")) : 60;
```
kalturaguy: i think that the mw.getConfig("cuePointsThreshold") should be called once, and not every time we call requestLiveCuepoints

PR #3794, opened 2018-04-25
fix(SUP-14173): Change screen orientation detection

ErezBouskila:
@eitanavgil Please review PR.
Thanks

### DIFF
--- a/modules/Quiz/resources/quiz.js
+++ b/modules/Quiz/resources/quiz.js
@@ -220,7 +220,7 @@
                 }
             });
             if (_this.getPlayer().getInterface().hasClass("mobile")){
-                $( window ).on( "orientationchange", function() {
+                $( window ).on( "orientationchange resize", function() {
                     if(_this.ivqShowScreenMode && _this.isPortrait() ) {
                         _this.showPortraitWarning();
                     }else {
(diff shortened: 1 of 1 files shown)


modules/Quiz/resources/quiz.js
```diff
@@ -689,7 +689,7 @@
         isPortrait: function () {
             if (
                 this.getPlayer().getInterface().hasClass("mobile")
-                && (screen.orientation.angle === 0 || screen.orientation.angle === 180)
```
eitanavgil: @ErezBouskila  
This looks OK but I am afraid that at some point one of the element in the chain parent.document.body.parentNode.clientWidth or parent.document.body.parentNode.clientHeight will be undefined. Perhaps in some embed types like iframe) 
Can we add a safer test ( if parent && parent.document .... )

eitanavgil:
@OrenMe 
@ErezBouskila  said he tested iframes in KMC and also outside - this is why I asked to add this over-protecting tests. U think we need to also add a try& catch?

OrenMe:
@eitanavgil there’s a method to check if you are in a friendly iframe or not. It should suffice. No need for try catch or lengthy checks.

ErezBouskila:
@eitanavgil @OrenMe Please review new fix.
Thanks

PR #3797, opened 2018-05-01
fix(SUP-14288): iOS auto captions not loaded on first launch

ErezBouskila:
@OrenMe Fix is intended for 2.70, please review.

### DIFF
--- a/modules/KalturaSupport/resources/mw.ClosedCaptions.js
+++ b/modules/KalturaSupport/resources/mw.ClosedCaptions.js
@@ -664,12 +664,20 @@
 		},
 		selectDefaultIosTrack: function (defaultLangKey) {
 			var _this = this;
-			this.once( 'playing', function (){
-				setTimeout(function () {
-					_this.log('selectDefaultIosTrack: ' + defaultLangKey);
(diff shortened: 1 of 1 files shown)


modules/KalturaSupport/resources/mw.ClosedCaptions.js
```diff
@@ -664,12 +664,20 @@
...
+				_this.selectTextTrack(defaultLangKey);
+			} else {
+				_this.once( 'playing', function (){
+					_this.selectTextTrack(defaultLangKey);
+				});
+			}
+		},
+		selectTextTrack: function(defaultLangKey) {
+			var _this = this;
+			setTimeout(function () {
```
OrenMe: @ErezBouskila better keep this timeout id in a var and if change media is done while it is still pending then clear it

yairans: @oren this is a timeout (one of million in V2...) not an interval. why to keep it on this?

yairans: LGTM

PR #3813, opened 2018-05-11
WEBC-1185-allow-anonymous-qa: change name to value from config

MaksymKorniiov:


modules/QnA/resources/qnaService.js
```diff
@@ -192,6 +192,9 @@
             else if (this.getType() === "AnswerOnAir"){
                 return gM('qna-answer-on-air');
             }
+            else if (this.cuePoint().overrideModeratorName !== false){
```
eitanavgil: @MaksymKorniiov  
I don't know why but this looks fishy. This function is some kind of a translation function and it handles content by 'this.getType()'. Your addition to this function doesn't look like the same syntax and I don't think it will reach the default  this.cuePoint().userId. Maybe we can write this in a different way?

modules/QnA/resources/qnaService.js
```diff
@@ -243,6 +246,7 @@
             // Setup player ref:
             this.embedPlayer = embedPlayer;
             this.qnaPlugin = qnaPlugin;
+            this.overrideModeratorName =  (qnaPlugin.getConfig('overrideModeratorName')) ? qnaPlugin.getConfig('overrideModeratorName') : false;
```
eitanavgil: why (qnaPlugin.getConfig('overrideModeratorName')) and not qnaPlugin.getConfig('overrideModeratorName')

PR #3817, opened 2018-05-18
FEV-182-submit-button:

MaksymKorniiov:
1. fix keyboard tab order of cp and submit button
2. add aria-label
3. fix click issue with continue button on "almost done" screen

### DIFF
--- a/modules/Tvpapi/resources/tvpapiAnalytics.js
+++ b/modules/Tvpapi/resources/tvpapiAnalytics.js
@@ -5,7 +5,8 @@
 
 		defaultConfig: {
 			mediaHitInterval: 30,
-			startTime:null
+			startTime:null,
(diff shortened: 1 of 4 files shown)


modules/Tvpapi/resources/tvpapiAnalytics.js
```diff
@@ -5,14 +5,16 @@
 
 		defaultConfig: {
 			mediaHitInterval: 30,
-			startTime:null
+			startTime:null,
```
eitanavgil: @MaksymKorniiov  how did this got in? 
Please fix this - it is not relevant to our ticket

MaksymKorniiov: @eitanavgil  this is not my commit

PR #3818, opened 2018-05-18
FEV-184-chapter-side-menu-buttons add correct tabindex and aria label

MaksymKorniiov:


modules/Tvpapi/resources/tvpapiAnalytics.js
```diff
@@ -5,14 +5,16 @@
 
 		defaultConfig: {
 			mediaHitInterval: 30,
-			startTime:null
+			startTime:null,
```
eitanavgil: Again - how is this relevant?

MaksymKorniiov: @eitanavgil this is not my commit

eitanavgil: @MaksymKorniiov  - I see that now. 
@chanakabr why did you commit this into our branch?

PR #3843, opened 2018-06-14
fix(sup-14444): addressed the matter with the hidden QnA div on VOD

shalommeoded:
@eitanavgil Please review my PR

### DIFF
--- a/modules/QnA/resources/qna.js
+++ b/modules/QnA/resources/qna.js
@@ -258,7 +258,7 @@
 						mw.log("failed to access window['parent'] for creating $qnaListContainer");
 					}
 				}
-				else{
+                if ( this.getPlayer().isLive() ) {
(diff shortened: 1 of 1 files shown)


modules/QnA/resources/qna.js
```diff
@@ -258,7 +258,7 @@
 						mw.log("failed to access window['parent'] for creating $qnaListContainer");
 					}
 				}
-				else{
+                if ( this.getPlayer().isLive() ) {
```
eitanavgil: @shalommeoded  
from here it seems that this code will be running also when this is rendered on-page. Is this what we want? 
if not let's change to else-if. 
Please make sure we check this with running players (onpage and not onpage) before we deliver this to QA

shalommeoded: @eitanavgil We also want the module not to show on-page when it is a VOD, since it will also just return a black div, so yes.

PR #3852, opened 2018-06-24
fix(SUP-13922):entry is playing audio only in edge when on embedframe

shalommeoded:
@OrenMe please review this PR


itaykinnrot:
I don't like this solution - i guess there is a reason for the element to be in this position state - will it effect the player? maybe we just need to remove and re-add it? lets discuss @shalommeoded

shalommeoded:
@itaykinnrot Hi, it seems that this issue is due to the 'Absolute' positioning in Edge,
However, it seems that the issue is resolved when positioning it to 'static' or 'none'.
What do you think?

PR #3861, opened 2018-07-02
fix(sup-14798): fixed the issue with chrome cast in Iframe

shalommeoded:
@OrenMe Please review my PR.

Fix:
1. DOM selector was unable to access the 'top' event.
2. Added the docContext.


shalommeoded:
@OrenMe We have fixed the matter with Chrome cast on unfriendly iFrame.

PR #3863, opened 2018-07-05
fix(sup-14444): the qna div should not be seen on live

shalommeoded:
@eitanavgil please review this PR


eitanavgil:
@shomronk  FYI - merged to 2.70. We can now tag this as a HF

PR #3868, opened 2018-07-15
fix(SUP-14895): addressed the matter with dual streams on Android

shalommeoded:
@OrenMe @eitanavgil Please review this PR

### DIFF
--- a/modules/KalturaSupport/resources/mw.KWidgetSupport.js
+++ b/modules/KalturaSupport/resources/mw.KWidgetSupport.js
@@ -1846,7 +1846,7 @@ mw.KWidgetSupport.prototype = {
 		if( !this.removedAdaptiveFlavors &&
 				(mw.isAndroid() && !mw.isNativeApp()) &&
 				hasH264Flavor &&
-				mw.getConfig( 'Kaltura.LeadHLSOnAndroid' ) == false ) {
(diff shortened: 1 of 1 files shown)


OrenMe:
@shalommeoded please explain the fix

shalommeoded:
@OrenMe Seems to be a regression from version 2.64 rc6 I have pointed this in the JIRA

OrenMe:
@shalommeoded please explain the fix itself, why is this change fixes the issue? It seems just the same, unless the value is not Boolean all the time and then !undefined or !null are correct

shalommeoded:
@OrenMe You are correct, just reinspected the Embedded HTML and it seems that there was an added "KalturaLeadHLSOnAndroid": false to the test flashvars, it seems that the correct fix will be to change the false parameter to null as the KalturaLeadHLSOnAndroid is never by default.

shalommeoded:
@OrenMe I have re-done the changes.

shalommeoded:
@OrenMe  
QA had returned this.
I can see in our API documentation that:

_+Kaltura.LeadHLSOnAndroid
If Apple HLS streams should be used when available on Android devices, by default progressive streams are used on Android because of Android HLS compatibility issues.+_

That this means that we should alter the  "Kaltura.LeadHLSOnAndroid" default settings to false?
As it seems to resolve the issue.

PR #3876, opened 2018-07-19
feat(FEC-8087): add absolutePosition parameter to Kava events

yairans:


erankor:
@yairans, I think it makes sense to also save a timestamp of 'the time you got the id3 tag' (can be just the local time of the user's computer), and if it's too old (>20 sec) - don't send it. Consider a case where for some reason you stop getting the id3 tags, in this case you will keep reporting the same timestamp indefinitely.

OrenMe:
@erankor should we set it to `null` if it was already reported with a certain timestamp or should we keep sending same time in multiple events?
In which case will id3 tags will stop arriving? is it expected that for a stream with id3 tags they will be available in some part of the stream and will not be available in another part of it?

PR #3877, opened 2018-07-22
fix(FEC-7700): set delivery type when adapter is chosen

OrenMe:


yairans:
@OrenMe Are you sure your change get along with https://github.com/kaltura/mwEmbed/blob/705544054bd4d2ae7dda67c6b86a9fee0d263641/modules/KalturaSupport/components/sourceSelector.js#L255 ?

OrenMe:
@yairans apparently... yes :-)

PR #3880, opened 2018-07-23
fix(SUP-14960): Moved the quarterly play events flags to the onChangeMedia function

shalommeoded:
@OrenMe Please review this PR.
Per our F2F and verifying the matter with Eran K I have removed the quarterly play events from resetting upon replaying\seeing the entry,
As a result the drop-off calculations of (#playThrough events / 4) / (#plays) should be corrected.

### DIFF
--- a/modules/KalturaSupport/resources/mw.kAnalony.js
+++ b/modules/KalturaSupport/resources/mw.kAnalony.js
@@ -128,6 +128,10 @@
 				_this.firstPlay = true;
                 _this.entryPlayCounter++;
                 _this.playSentOnStart = false;
+                _this._p25Once = false;
+                _this._p50Once = false;
+                _this._p75Once = false;
+                _this._p100Once = false;
 			});
 
 			this.embedPlayer.bindHelper( 'userInitiatedPlay' , function () {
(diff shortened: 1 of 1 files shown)


modules/KalturaSupport/resources/mw.kAnalony.js
```diff
@@ -314,10 +314,6 @@
 			});
 		},
 		resetPlayerflags:function(){
-			this._p25Once = false;
```
OrenMe: @shalommeoded now you never reset these flags, so in change media you won't report them again

shalommeoded: @OrenMe While inspecting the fix on Playlist and it changed media, and the quarterly flags were showing as expected.

It seems that the session is restarting when the 'onChangeMeida' is called:
https://github.com/shalommeoded/mwEmbed/blob/bf98f379e4902ae3919d547ff9323781980b1c26/modules/KalturaSupport/resources/mw.kAnalony.js#L116

Please correct me if I have misunderstood your comment.

OrenMe: reset session doesn't clear the percentiles flags. Moreover you just delted the only code that resets these flags so they won't be reset at all.

shalommeoded: @OrenMe So instead of deleting the flags, I have moved them into the onChangeMedia function, which will reset the flags to flase again with the onChangeMedia is called.

PR #3883, opened 2018-07-26
fix(SUP-15010, SUP-15012): Addressed the matter with files returning 0 bytes when downloaded

shalommeoded:
@OrenMe @itaykinnrot 
Per my discussion with Core devs (Moshe Maor, Erez Gotlib) they have provided me with this fix, it seems that we were missing the following code on the player.

### DIFF
--- a/modules/KalturaSupport/download.php
+++ b/modules/KalturaSupport/download.php
@@ -86,10 +86,10 @@ class downloadEntry {
 				header("Location: " . $flavorUrl );
 
 			}else{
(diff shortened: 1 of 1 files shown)


OrenMe:
@shalommeoded what is the issue? How does this code fix it?

shalommeoded:
@erankor please review this PR

Per our f2f I have added the restore parameters to the else statement.

erankor:
@shalommeoded, looks ok, other than the indent - please change to tabs to match the rest of the code

PR #3884, opened 2018-07-30
FEC 8433 v2 kanalony referrer is double escaped

shalommeoded:
@OrenMe Please review this PR.

I removed the encodeURI functions from the Kanalony code

### DIFF
--- a/modules/KalturaSupport/resources/mw.kAnalony.js
+++ b/modules/KalturaSupport/resources/mw.kAnalony.js
@@ -550,17 +550,16 @@
             var pageReferrer =  statsEvent[ 'referrer' ];
             var queryPos = pageReferrer.indexOf("?");
             if (queryPos > 0) {
-                pageReferrer = pageReferrer.substring(0, queryPos);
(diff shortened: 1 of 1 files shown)


modules/KalturaSupport/resources/mw.kAnalony.js
```diff
@@ -550,17 +550,16 @@
             var pageReferrer =  statsEvent[ 'referrer' ];
             var queryPos = pageReferrer.indexOf("?");
             if (queryPos > 0) {
-                pageReferrer = pageReferrer.substring(0, queryPos);
+              pageReferrer = pageReferrer.substring(0, queryPos);
             }
 
-            var encodedReferrer = encodeURIComponent(pageReferrer);
```
OrenMe: @shalommeoded the first test was OK, you need to test it with `encodeURIComponent`

shalommeoded: @OrenMe I have re-added it.

PR #3895, opened 2018-08-18
fix(FEC-8485): fix iOS text track selection and user settings

OrenMe:


modules/KalturaSupport/resources/mw.ClosedCaptions.js
```diff
@@ -159,6 +160,26 @@
...
 			}
 
+            /**
+			 * support saving user selected text track language in iOS native player
+             */
+			function maybeRegisterTextTrackChangeHandler () {
+                if( this.isNativeIOSPlayback() && this.getConfig('useCookie') && !this._registeredNativeTrackChangeHandler){
+                    this._registeredNativeTrackChangeHandler = true;
+                    this.embedPlayer.getPlayerElement().textTracks.addEventListener("change", function(){
+                        var activeSubtitle = this.embedPlayer.getActiveSubtitle();
```
yairans: doesn't work!!!
`this` ref is wrong. should be `_this`

yairans: doesn't work!!!
`this` ref is wrong. should be `_this`

modules/KalturaSupport/resources/mw.ClosedCaptions.js
```diff
@@ -159,6 +160,26 @@
...
+                        var activeLanguage = null;
+                        if (activeSubtitle) {
+                            activeLanguage = activeLanguage.language;
```
yairans: `null` exception. `activeLanguage` is `null` 2 lines above....

PR #3904, opened 2018-09-13
fix(SUP-15427): loadThumbnailWithKs not applying to all thumbnail use-cases.

shalommeoded:
@eitanavgil Per our discussion, we saw that when the loadThumbnailWithKs flashvar applies only to the player "Poster" and not on the other Thumbnail use-cases, for example, Chapters cue point thumbnails.

Please review this PR.


eitanavgil:
@shalommeoded  make sure this also gets into 2.72 once it will be available. I am not sure that we will release a HF for this fix.

PR #3911, opened 2018-10-02
fix(SUP-15625): Fix the chapters plugin right positioning issue.

shalommeoded:
@eitanavgil Please review this PR.


shalommeoded:
@eitanavgil 
Per our discussion, I have changed the CSS:

1. Removed the redundant Class and added them to the left positioning classes.
2. Removed the text-decoration of the right side as well.

PR #3912, opened 2018-10-03
fix(SUP-15645): play event is counted twice in KAVA once a pre-roll is triggered first

shalommeoded:
@OrenMe 
This issue was a regression bug of  FEC-8456.
Fix:
1. Added the onPlayStatus and set it to false on the Default config.
2. And added it to the PLAY count logic.
3. Altered the RESUME count logic to trigger when the currentState is not "PLAY" that to avoid any miscounts.

### DIFF
--- a/modules/KalturaSupport/resources/mw.kAnalony.js
+++ b/modules/KalturaSupport/resources/mw.kAnalony.js
@@ -60,7 +60,7 @@
         playSentOnStart: false,
 		absolutePosition: null,
 		id3TagEventTime: null,
-
+		onPlayStatus: false,
 		smartSetInterval:function(callback,time,monitorObj) {
 			var _this = this;
 			//create the timer speed, a counter and a starting timestamp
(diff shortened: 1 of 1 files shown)


modules/KalturaSupport/resources/mw.kAnalony.js
```diff
@@ -144,13 +144,14 @@
                 }
 
 				if ( !this.isInSequence() && (_this.firstPlay || _this.embedPlayer.currentState !== "play") ){
-					if ( _this.firstPlay ){
+					if ( _this.firstPlay && !_this.onPlayStatus ) {
+						_this.onPlayStatus = true;
```
OrenMe: need to reset the flag state on change media otherwise it will never enter here after first time

shalommeoded: @OrenMe Added it to the 'onChangeMedia' function.

PR #3922, opened 2018-10-15
Fix(FEC-8545): Fixed the Seek issue in the IOS Inline player + added a userAgent for IOS12

shalommeoded:
@OrenMe Plese review this PR.

1. I have added the isIOSAbove7 and isIOS12 to the if condition, as it was unable to recognize the readyState element (0 = empty) therefore, was never on "ready".

2. Added a isIOS12 userAgent.

### DIFF
--- a/modules/EmbedPlayer/resources/mw.EmbedPlayerNative.js
+++ b/modules/EmbedPlayer/resources/mw.EmbedPlayerNative.js
@@ -481,7 +481,7 @@
 			// some initial calls to prime the seek:
 			if ( ( vid.currentTime === 0 && callbackCount === 0 ) && vid.readyState === 0 ) { //load video again if not loaded yet (vid.readyState === 0)
 				// when seeking turn off preload none and issue a load call.
-				if(mw.isIpad()){
+				if(mw.isIpad() || mw.isIOSAbove7() || mw.isIOS12()){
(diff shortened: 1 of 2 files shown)


modules/EmbedPlayer/resources/mw.EmbedPlayerNative.js
```diff
@@ -481,7 +481,7 @@
 			// some initial calls to prime the seek:
 			if ( ( vid.currentTime === 0 && callbackCount === 0 ) && vid.readyState === 0 ) { //load video again if not loaded yet (vid.readyState === 0)
 				// when seeking turn off preload none and issue a load call.
-				if(mw.isIpad()){
+				if(mw.isIpad() || mw.isIOSAbove7() || mw.isIOS12()){
```
OrenMe: We already have an example of getting safari above a specific version. It doesn’t make any sense to add ios12 like this.
Either patch the iOS above 7 to include 12 and anything above, or just change it to  isIos as I don’t believe there are too many devices that use ios6 and below if at all - and you can verify this in the Kava analytics dashboard.

shalommeoded:
@OrenMe 
Per our F2F, it seems that the code wasn't updated correctly.
However, it seems that now when seeking the player will restart the entry upon finishing the ads, however, I do think that this issue is related to the known issue with IMA SDK bug in IOS.

PR #3928, opened 2018-11-01
2.72 FEV-232	

kalturaguy:


kalturaguy:
The purpose of this PR is to pass ONLY playManifest and thumbAsset->getUrl via eCDN proxy API and the rest (analytics, other API calls) via CDN API.

OrenMe:
@kalturaguy please delete your branch after you merge it! Thanks.

PR #3929, opened 2018-11-01
fix (SUP-15880/15796): AutoMute doesn't work with Autoplay in chrome

shalommeoded:
@OrenMe 

Per our f2f, I was able to see that the toggleMute was called out too late due to the setTimeout function.

Lowering it the wait time to 0 seems to fix the issue.


modules/KalturaSupport/resources/mw.KWidgetSupport.js
```diff
@@ -951,7 +951,7 @@ mw.KWidgetSupport.prototype = {
 		if( autoMute && !mw.isMobileDevice()){
 			setTimeout(function(){
 				embedPlayer.toggleMute( true );
-			},300);
+			},0);
```
OrenMe: need to understand why this was 300 in the first place before changing it...

PR #3930, opened 2018-11-04
An alternative implementation to the original RaptMedia plugin

jessp01:
Unlike the original, the `RaptMediaNg` doesn't make use of stitched playlists to serve the content but rather, loads each entry in the playlist individually, thus avoiding the need to compensate for uncertainty, see:
https://github.com/kaltura/mwEmbed/blob/master/modules/RaptMedia/resources/mw.RaptMedia.js#L385-L408

The implementation in modules/RaptMediaNg/resources/mw.RaptMediaNg.js was taken from https://github.com/FlixMaster/mwEmbed/blob/rapt-media-change-media/modules/RaptMedia/resources/mw.RaptMedia.js [made against the v2.68 mwembed branch] and all credit for that should go to Hargobind S. Khalsa [@khalsah].

Based on the above implementation, I created the new `RaptMediaNg` plugin. This was done in order to enable specific customers whose playlists [AKA: Rapt projects], severely malfunction with the existing plugin, to consume this alternative plugin while ensuring that no other customer loads it by accident.

The code for the following supporting plugins:
- RaptMediaNgDurationLabel
- RaptMediaNgPlayPauseBtn

is aligned with that of the originals in 2.73 branch [https://github.com/kaltura/mwEmbed/tree/2.73/modules/RaptMedia/resources].

Below are the tests I performed thus far:
- Ensured that the new plugin [`RaptMediaNg`] and the original [`RaptMedia`] can co-exist and that there are no conflicts
- Made sure the RaptMediaNg/tests/RaptMediaNg.qunit.html works as expected
- Tested the new implementation with the following browsers:
  - Firefox [latest - 63.0]: Linux and Windows [tested by Zoe]
  - Chrome: Linux [Chromium 70.0.3538], Windows [tested by Zoe] and mobile [iPhone + Chrome - Zoe, Android + Chrome - tested by me] 
  - Zoe Fiddes, the customer's CSM also tested on IE/Edge

An embed code that makes use of the new plugin can be found here:
https://demo.kaltura.me/rapt3/modules/RaptMediaNg/tests/zoe1.html

The same embed code but with the original plugin, can be found here:
https://demo.kaltura.me/rapt3/modules/RaptMedia/tests/zoe1.html


jessp01:
@eitanavgil, @khalsah - I'd really appreciate it if you could give this a look.

modules/RaptMediaNg/RaptMediaNg.json
```diff
@@ -0,0 +1,34 @@
+{
+	"mw.RaptMediaNg": {
+		"scripts": [
+			"resources/engine.min.js",
```
eitanavgil: if this exists on the previous rapt implementation - we can use the same file and not duplicate it

modules/RaptMediaNg/RaptMediaNg.json
```diff
@@ -0,0 +1,34 @@
...
+			"resources/mw.RaptMediaNg.js"
+		],
+		"styles": "resources/RaptMediaNgPlugin.css",
```
eitanavgil: if (not sure) this is the same CSS as prev version - we can use it too.

modules/RaptMediaNg/RaptMediaNg.json
```diff
@@ -0,0 +1,34 @@
...
+		"dependencies": [
+			"mw.KBaseComponent"
+		],
+		"kalturaPluginName": "raptMediaNg"
+	},
+	"mw.RaptMediaNgDurationLabel": {
```
eitanavgil: separating the stitched to playlist - do we need that label and alternative scrubber?

modules/RaptMediaNg/RaptMediaNg.json
jessp01: Hi @eitanavgil,

Thank you for taking the time to review.
Technically speaking, we could use resources/engine.min.js, resources/RaptMediaNgPlugin.css and resources/mw.RaptMediaNgDurationLabel.js as they exist in the original `RaptMedia` plugin [modules/RaptMedia/resources]. The reason why I duplicated these [and renamed them accordingly] is to make the new plugin as independent/self contained as possible so that, if in the future, changes are made to the original, they will not affect the new one and, by the same token, should the new plugin necessitate changes in these files for whatever reason, they will not affect the original. Does that make sense?

modules/RaptMediaNg/resources/mw.RaptMediaNg.js
```diff
@@ -0,0 +1,508 @@
...
+			this.setConfig('info', undefined, true);
+
+			this.pendingEntryId = null;
+			this.targetEntryId = null;
+		},
+
+		addBindings: function() {
+			var _this = this;
+
+			this.bind('raptMediaNg_doPlay', function(event) {
```
khalsah: Since these bindings (`raptMedia_doPlay`, `raptMedia_doPause`, etc) are primarily designed to support custom external API usage, I would recommend leaving the prefix as `raptMedia_` instead of changing it to `raptMediaNg_`. That way it will be easier to switch between the two plugins without having to update any custom API usage the user may have external to the player.

I do realize this creates some room for confusion, particularly since there are a few events that are only indented to be internal. As such I would probably recommend sticking with a `raptMedia_` prefix for all events.

modules/RaptMediaNg/resources/mw.RaptMediaNg.js
```diff
@@ -0,0 +1,508 @@
+/**
+ * The RaptMediaNg plugin integrates the RaptMedia Engine to the Kaltura Player.
+ * RaptMedia adds clickable interactive layer that accompanies your video content and can do things like:
+ * cue or launch different media plays, jump to specific timecode, trigger an event on your webpage and launch a new web page or an app.
+ * This plugin also makes use of accompanying plugin RaptMediaNgScrubber plugin to override the default scrubber behavior to fit a Rapt Media experience.
+ * With RaptMediaNgScrubber plugin the scrubber can interact within the context of a single RaptMedia clip instead of just the entire stitched playlist.
+ * The RaptMediaNg plugin integrates the RaptMedia Engine to the Kaltura Player.
+ * It also makes use of accompanying plugin RaptMediaNgDurationLabel used to override the default player DurationLabel to behave according to the RaptMedia Sequence rather than show the overall playlist duration.
```
khalsah: This comment is not accurate as the "other" plugins such as `RaptMediaNgScrubber`  and `RaptMediaNgDurationLabel` aren't required in this version of the plugin and, essentially, don't _really_ do anything.

modules/RaptMediaNg/resources/mw.RaptMediaNg.js
```diff
@@ -0,0 +1,508 @@
...
+				// KalturaSupport_EntryDataReady can be called synchronously from a
+				// `checkPlayerSourcesEvent` handler if the required data is already
+				// cached. In that case `_this.playbackCallback` may not be available
+				// synchronously, so we force asynchronous evaluation
+				setTimeout(function() {
+					_this.log('Checking if Entry is an Interactive Video');
+
+					var raptProjectId = _this.readRaptProjectId();
+					if (raptProjectId) {
+						_this.once('raptMediaNg_ready', _this.playbackCallback);
```
khalsah: This is another event designed for external consumption, as such I would consider leaving it with the `raptMedia_` prefix.

modules/RaptMediaNg/resources/mw.RaptMediaNg.js
```diff
@@ -0,0 +1,508 @@
...
+					return _this.reject(new AbortError);
+				}
+
+				if (_this.$el)
+					_this.$el.show();
+
+				_this.log('Starting rapt project');
+
+				_this.setConfig('status', 'enabled', true);
+				_this.emit('raptMediaNg_ready');
```
khalsah: Same comment as other `raptMediaNg_`  usage.

modules/RaptMediaNg/RaptMediaNg.json
khalsah: The CSS has changed slightly and so can't be de-duplicated.

modules/RaptMediaNg/RaptMediaNg.json
khalsah: I'm going to concur here. Other than the "core" `mw.RaptMediaNg` plugin all the others are simply shims that were required to patch bits of the UI to hide the fact that it was one large stitched playlist. As such is should be possible to completely remove them without any adverse effect.

I do see at least an argument for keeping them to preserve a closer to 1-to-1 mapping and minimize confusion when switching between the two different versions.

modules/RaptMediaNg/resources/mw.RaptMediaNg.js
jessp01: Hi @khalsah - 

First of all, thanks for reviewing.
Generally speaking, I concur [it will be hard for anyone to argue with this logic, IMHO].
However, because of the nature of this particular delivery, I'd rather keep things as they are for now. If we end up providing this to additional customers and the plugin is recognised as an officially supported one, I will be glad to do some re-factoring [shouldn't take more than 5 minutes] and let it run through a proper QA process.

The above true to all your other comments as well.

OrenMe:
@jessp01 are you planning to merge this? can you please conclude if this is to be merged or closed.

OrenMe:
stale PR.

PR #3931, opened 2018-11-05
fix(SUP-15880): Addressed the issue when the AutoPlay doesn't work with AutoMute

shalommeoded:
@OrenMe please review this PR.

Per our discussion, we just implemented this in here to bypass the setTimeout that is set in mw.KWidgetSupport:

https://github.com/shalommeoded/mwEmbed/blob/5fe2421d638a7f45e91ee07cfe6da72e47310415/modules/KalturaSupport/resources/mw.KWidgetSupport.js#L952

### DIFF
--- a/modules/EmbedPlayer/resources/mw.EmbedPlayerNative.js
+++ b/modules/EmbedPlayer/resources/mw.EmbedPlayerNative.js
@@ -127,6 +127,9 @@
                 });
             } else {
                 this.toggleMute( true );
+                if (mw.isMobileDevice()) {
+                	this.mobileAutoPlay = true;
+              	}
             }
 
 			this.addBindings();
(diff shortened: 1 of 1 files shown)


shalommeoded:
@OrenMe 
So it seems that the change will not work in mobile(tested this again.)
However, I did add this condition for mobile, and it seems to work:
 
`if (this.shouldAutoPlayMuted()) {
    this.autoplay = true;
    if (!mw.isIphone() && this.inline) {
        this.inline = false;
        $(this.getPlayerElement()).removeAttr('playsinline');
    }
    this.mobileAutoPlay = true;
    this.bindOnceHelper('playerReady', function () {
        _this.setVolume(0);
    });
} else {
    this.toggleMute( true );
      if (mw.isMobileDevice()) {
        this.mobileAutoPlay = true;
      }
}`
**BTW:**

It doesn’t seem that MobileAutoPlay and AutoMute ever worked before on Mobile (I tested with FireFox and Chrome in Android an IOS as well) and it seems that this use-case wasn’t supported at all, at least from what I saw while downgrading the player to till version 2.62.

PR #3933, opened 2018-11-06
fix(SUP-15798): live entries are showing black bars upon exiting full screen

shalommeoded:
@OrenMe 
Please review this PR.

The issue is a regression of SUP-4030:
https://github.com/kaltura/mwEmbed/commit/575ee85d4e03acc003d4920b374265b0bd03f052

I just added a condition that will only trigger when the scrubber is hidden (when DVR is set to false).


shalommeoded:
Thank you I will suggest that to the client instead.

PR #3936, opened 2018-11-10
remove legacy live analytics

kalturaguy:


kalturaguy:
@OrenMe  @erankor   This is not used anymore....

OrenMe:
@kalturaguy @erankor so this is approved for removal?

DBezemer:
@OrenMe this is still used in OnPrem as Kava is not ready for OnPrem yet. Can we keep it in the player for the time being (maybe just disabled by default). Otherwise we're stuck with v2.72 until Kava is released for Onprem

kalturaguy:
@DBezemer  Do you have "legacy" live analytics? Kanalony?

PR #3938, opened 2018-11-20
fix(SUP-16258): Issue with PlaybackSelector in mobile.

shalommeoded:
Change the condition of the if statement in order to avoid entering the "handlePlayerInstanceUpdate" function and just enter "updatePlaybackRate".


OrenMe:
please use the config option for this. I think it's not necessary to change the default due the affected regression risk.

PR #3942, opened 2018-11-25
[Player_V2][KAVA] - In mobile device, locking the phone does not cause a pause event (33)

dan-ziv:


modules/KalturaSupport/resources/mw.kAnalony.js
```diff
@@ -192,6 +192,10 @@
                 if ( _this.embedPlayer.isDVR() ) {
 					_this.dvr = true;
 				}
+                if (!_this.timer.isStopped()){
+                    _this.timer.stop();
```
OrenMe: @dan-ziv doesn't make a lot of sense to send PAUSE on seeking - ?

PR #3944, opened 2018-11-26
Fec 8608 ios autoplay is not working with playsinline=false configuration

shalommeoded:
Reverted back to the original function, as the changes didn't work with playsinline=false configurations.


OrenMe:
@shalommeoded can we close this?

PR #3946, opened 2018-12-04
fix: unmute doesn't get destroyed in playlist player

OrenMe:
destroy super does unbind and also remove so no need to implement the method, which causes the bug in playlist player anyway where the indication doesn't go away after user clicks on playlist

### DIFF
--- a/modules/unMuteOverlayButton/resources/unMuteOverlayButton.js
+++ b/modules/unMuteOverlayButton/resources/unMuteOverlayButton.js
@@ -56,12 +56,6 @@
                 this.getComponent().fadeOut('slow');
             },
 
-            destroy: function () {
-                this.hide();
-                this.unbind('playerReady');
-                this.unbind('volumeChanged');
-            },
-
             getComponent: function () {
                 if (!this.$el) {
                     this.$el = $('<button/>')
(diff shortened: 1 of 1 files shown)


shalommeoded:
@OrenMe @dan-ziv  Looks good and works.

Just a side note in regards to the UI, it seems that the button is disappearing to abruptly, do you think that we should keep the "fadeOut" method? in order to make the button disappear in a more "elegant" way.

Maybe something like maybe:
`this.getComponent().fadeOut('slow', function() { $(this).remove(); });`

OrenMe:
@shalommeoded sounds good, feel free to check if it ok and you can suggest a code change on this PR itself: https://help.github.com/articles/incorporating-feedback-in-your-pull-request/#applying-a-suggested-change

modules/unMuteOverlayButton/resources/unMuteOverlayButton.js
```diff
@@ -56,12 +56,6 @@
                 this.getComponent().fadeOut('slow');
```
shalommeoded: To apply this to the hide function and keep the destroy function.
```suggestion
                hide: function () {
                this.getComponent().fadeOut('slow', function () {this.remove()});
            },

            destroy: function () {
                this.hide();
                this.unbind('playerReady');
                this.unbind('volumeChanged');
            },
```

kaltura-hooks:
Hi @OrenMe,

shalommeoded:
@OrenMe 
Please check the new commit.

PR #3951, opened 2018-12-19
SUP-16617 - thumbnails are not showing in IOS

shalommeoded:
The canAutoPlay function is still trying to autoPlay in the Native IOS Player however fails due to the player not being inline, that is why the thumbnails are not showing when the autoPlay and mobileAutoPlay are set to true.

The fix is to add the original return autoplay without the isMuted() function to the if isiPhone validation and I have also added && mw.getConfig("EmbedPlayer.WebKitPlaysInline") !== true ) for safety mesurs.

### DIFF
--- a/modules/EmbedPlayer/resources/mw.EmbedPlayerNative.js
+++ b/modules/EmbedPlayer/resources/mw.EmbedPlayerNative.js
@@ -333,11 +333,10 @@
 		canAutoPlay: function () {
 			if ( mw.isMobileDevice() ) {
 				var playsinline = true;
-				if (( mw.isIphone() && mw.getConfig("EmbedPlayer.WebKitPlaysInline") !== true )) {
+				if ( mw.isIphone()) {
 					playsinline = this.inline;
-					return (this.mobileAutoPlay && playsinline) || this.mobilePlayed;
 				}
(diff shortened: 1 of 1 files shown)


shalommeoded:
@OrenMe 
I'm aware that 2.73 is closed, we need to push this fix to 2.74.

shalommeoded:
@OrenMe 
Changed and also tested this on iPhone, Andriod and, Desktop.

modules/EmbedPlayer/resources/mw.EmbedPlayerNative.js
```diff
@@ -333,10 +333,10 @@
 		canAutoPlay: function () {
 			if ( mw.isMobileDevice() ) {
 				var playsinline = true;
-				if ( mw.isIphone() ) {
+				if ( mw.isIphone()) {
 					playsinline = this.inline;
 				}
-				return (this.mobileAutoPlay && playsinline) || this.mobilePlayed || this.isMuted();
+				return (this.mobileAutoPlay || this.isMuted()) && playsinline || this.mobilePlayed;
```
OrenMe:
```suggestion
				return ((this.mobileAutoPlay || this.isMuted()) && playsinline) || this.mobilePlayed;
```

OrenMe:
@shalommeoded please make sure to base your PRs to 2.74 branch

PR #3952, opened 2019-01-07
FEV-236 improved analytics

kalturaguy:


modules/KalturaSupport/resources/mw.kAnalony.js
```diff
@@ -104,6 +114,16 @@
 		},
 
 		setup: function( ) {
+			// grab user uuid from cookie or get one and save it to the cookies of this domain
+			if(mw.getConfig("forceUUID")){
```
kalturaguy: maybe let's call it a friendly name: "useBrowserIdAsUserId"

modules/KalturaSupport/resources/mw.kAnalony.js
```diff
@@ -292,8 +312,45 @@
...
+			});
+
+			this.embedPlayer.bindHelper('hlsManifestLoadedWithStats', function(e,data){
+			    _this.stats.manifestkDownloadTime= (data.stats.tload-data.stats.trequest).toFixed(2);
+			});
+
+
+			var hlsFunc = function(e, data){
+				// this is hls 
+				_this.stats.playbackEngine = "hlsjs";
```
kalturaguy: maybe we should unbind here?

modules/KalturaSupport/resources/mw.kAnalony.js
```diff
@@ -433,30 +490,71 @@
...
 				if ( !_this.monitorViewEvents ){
 					_this.stopViewTracking();
 				}
 			},_this.reportingInterval,_this.monitorIntervalObj);
 
 		},
 
+		setCookie:function(name,value,days) {
+			var expires = "";
+			if (days) {
```
kalturaguy: 1.we never  pass expiry, so i think it will expire after refresh
2.with new cookie policy, i think we will have hard problem with cookie expiry as they now browser doesn
t support unlimited expiry
3. i think we need to consider localStorage, and if we are IE10 not to enable this feature and not send userId

modules/KalturaSupport/resources/mw.kAnalony.js
```diff
@@ -66,6 +66,16 @@
 		onPlayStatus: false,
         firstPlayRequestTime: null,
 
+        stats: {
+			chunksDownloaded: 0,
+			maxChunkDownloadTime:-1,
+			playbackEngine:-1,
+			manifestkDownloadTime: -1 ,
```
kalturaguy: there is mispelling manifest**k**DownloadTime

PR #3953, opened 2019-01-14
fix(SUP-16728): Can't download an entry with only a source.

shalommeoded:
the fix that will allow the player to download the source file without defaulting the **flavorParamsId** to 0.


modules/KalturaSupport/resources/uiConfComponents/downloadPlugin.js
```diff
@@ -30,15 +30,18 @@
 				downloadUrl += this.getPlayer().kwidgetid + '/uiconf_id/' + this.getPlayer().kuiconfid;
 				downloadUrl += '/entry_id/' + this.getPlayer().kentryid + '?forceDownload=true';
 				downloadUrl += '&downloadName=' + encodeURIComponent(this.getConfig('downloadName'));
-				if( this.getConfig('flavorParamsId') ){
+				if( this.getConfig('flavorParamsId') && !this.isSourceOnly()){
```
OrenMe: just put `if (!this.isSourceOnly())` around all three `if` statements 
so 
```javascript
if (this.isSourceOnly()) {
  downloadUrl += '&flavorParamsId=0'
} else {
  //all three other options here
}
```

PR #3954, opened 2019-01-20
fix(SUP-16764): Quiz.plugin pushes the captions upwards.

shalommeoded:
@eitanavgil 
Seems that the quiz CSS was pushing the captions upwards by 9vh for some reason.
I have just deleted the _!important_ since it was bothering the user in the use case when the quiz plugin is activated however the entry is not a quiz type.

### DIFF
--- a/modules/Quiz/resources/css/quiz.css
+++ b/modules/Quiz/resources/css/quiz.css
@@ -806,7 +806,7 @@ ol.second-row{
 }
 
 .track {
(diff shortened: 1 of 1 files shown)


shalommeoded:
@eitanavgil 
Well, it seems that the issue is with regular entries with captions, the Quiz CSS is taking effect even if the entry is not a quiz type.

PR #3976, opened 2019-03-05
add copyright and term of use violation to moderation plugin

amirch1:


amirch1:
@OrenMe  - Please verify in test page and Studio as I don't have the test environment set

shalommeoded:
@amirch1 
I have tested the pull, and it workes on a test page, and I can see the added fields on the moderation options.

However, when flagging I am unable to see the reason for the "report" on the moderation tab on the KMC, it shows me that the reason is "flag 1", I assume this is due to the fact that these settings were added to the KMC yet.

amirch1:
@shalommeoded - good call! Seems KMC support should be added here: https://github.com/kaltura/kmc-ng/blob/0c56c6278c94f9558b741a1c036267ae97302800/src/applications/content-moderation-app/pipes/flagType.pipe.ts#L13-L26
For now, let's keep this PR open until we can conclude if we want to add this support to KMC. Thanks

OrenMe:
@amirch1 @shalommeoded what is the status of this? is the KMS change scheduled?

amirch1:
@OrenMe - this requires a fix in KMC as well. Should be prioritized by Lior B.

PR #3977, opened 2019-03-05
fix(PSVAMB-6336) 2nd video plays after 1st has been seeked

vladm3:
#### _PSVAMB-6336_: dualScreen player when paused, the secondary video continues to play
* HLS.js should be treated as a player that supports 'optimistic' seek
* Allow `mediaGroupSync` if player is stopped and seeking


modules/KalturaSupport/components/dualScreen/nativePlayer.js
```diff
@@ -85,7 +85,7 @@
         },
 
         supportsOptimisticSeeking: function supportsOptimisticSeeking() {
-            return !this.isABR();
+            return true;
```
eitanavgil: @vladm3  
Whats the point having a function that always returns true?

vladm3: @eitanavgil - `nativePlayer.js` and `flashPlayer.js` both inherit from `videoPlayer.js` and this method returns `false` for `flashPlayer.js`

PR #3989, opened 2019-03-26
fix paste and cut with their events

adimachmali:


adimachmali:
Description: QnA char counter feature: fix char count when using copy paste using their matching events

PR #4005, opened 2019-05-07
2.75 webc 1444 fix unread counter when announcement is deleted

adimachmali:


modules/QnA/resources/qnaModule.js
```diff
@@ -187,7 +187,8 @@
                 var count = 0;
                 ko.utils.arrayForEach(_this.myObservableArray(), function (thread) {
                     for (var i = 0; i < thread().entries().length; i++) {
-                        if (thread().entries()[i]().getType() !== 'Question' && !thread().entries()[i]().isRead()) {
+                        if (thread().entries()[i]().getType() !== 'Question' && !thread().entries()[i]().isRead()
```
OrenMe: would suggest doing:
```javascript
var entry = thread().entries()[i]();
if (entry.getType() !== 'Question' && !entry.isRead())
    && (entry..getType() === 'Announcement' ? entry.cuePoint().metadata.State !== 'Deleted' : true)) {   
```

Also, there seem to be a redundant logic here. If you check explicitly for Announcement type, why do you need to check for not a question type?
also, mixing all of this in the if condition with a ternary expression is overly complex expression anyway. Can we do better?

modules/QnA/resources/qnaModule.js
```diff
@@ -187,7 +187,8 @@
                 var count = 0;
                 ko.utils.arrayForEach(_this.myObservableArray(), function (thread) {
                     for (var i = 0; i < thread().entries().length; i++) {
-                        if (thread().entries()[i]().getType() !== 'Question' && !thread().entries()[i]().isRead()) {
+                        var entry = thread().entries()[i]();
+                        if (entry.getType() !== 'Question' && !entry.isRead() && !(entry.getType() === 'Announcement' && entry.cuePoint().metadata.State === 'Deleted')) {
                             count++;
                         }
                     }
```
adimachmali: we can write: 
if (entry.getType() === 'Answer' && !entry.isRead() || (entry.getType() === 'Announcement' && entry.cuePoint().metadata.State !== 'Deleted')) {
                            count++;
                        }

but I not sure if we want to this and enforce other types like that (maybe we want more types to be included)

OrenMe: Your choice, but I found it to be very complex and hard to understand.
It's better now but I don't make much sense of it when reading it

PR #4006, opened 2019-05-08
fix Css style - submit button z-indexing

adimachmali:


adimachmali:
Didn't know exactly if the PR should e against 2.75 or not but you will know how to release that.

PR #4007, opened 2019-05-12
fix(SUP 17081): undefined sources are added to sourceSelector menu as empty spaces

shalommeoded:
In some use cases, streamSelectors and ClosedCaption are rendering empty sources (undefined) as small empty spaces on the drop-down menu.

### DIFF
--- a/modules/KalturaSupport/components/sourceSelector.js
+++ b/modules/KalturaSupport/components/sourceSelector.js
@@ -142,9 +142,8 @@
                 this.getPlayer().mediaElement.removeSourceFlavor(sources);
             }
 
-			if( sources.length == 1 ){
-				// no need to do building menu logic. 
-				this.addSourceToMenu( sources[0], _this.getSourceTitle(sources[0]) );
+			if( sources.length == 1 && this.getPlayer().streamerType != "http" ){
+				// no need to do building menu logic.
 				return ;
 			}
 			// sort by height then bitrate:
--- a/modules/KalturaSupport/components/sourceSelector.js
+++ b/modules/KalturaSupport/components/sourceSelector.js
@@ -312,7 +312,7 @@
 			var _this = this;
 
             var sourceLabel = this.getSourceTitle( source );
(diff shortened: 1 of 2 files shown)


modules/KalturaSupport/components/sourceSelector.js
```diff
@@ -312,7 +312,7 @@
 			var _this = this;
 
             var sourceLabel = this.getSourceTitle( source );
-            if( $.inArray(sourceLabel, this.sourcesList) == -1 ) {
+            if( $.inArray(sourceLabel, this.sourcesList) == -1 && source.src !== undefined ) {
```
OrenMe: doesn't make a lot of sense - the issue is that we show the menu when there's adaptive stream with only one flavor.
We need to add a check for this and remove menu in this case.

modules/KalturaSupport/resources/mw.ClosedCaptions.js
```diff
@@ -1151,8 +1151,10 @@
             // Add text sources
             for (var j = 0; j < sources.length; j++) {
                 var src = sources[j];
-                this.addSourceButton(src);
-                items.push({'label': src.label, 'value': src.label});
+                if (src.label !== "") {
```
OrenMe: need to check why we get this source in Safari for this stream.
It has an empty captions text track.

modules/KalturaSupport/components/sourceSelector.js
```diff
@@ -142,9 +142,8 @@
                 this.getPlayer().mediaElement.removeSourceFlavor(sources);
             }
 
-			if( sources.length == 1 ){
-				// no need to do building menu logic. 
-				this.addSourceToMenu( sources[0], _this.getSourceTitle(sources[0]) );
+			if( sources.length == 1 && this.getPlayer().streamerType != "http" ){
```
OrenMe: Please explain the condition

shalommeoded: Sure, if the source length is 1 and if it's adaptive. 
If we are on progressive we don't add the "Auto" tag to he menu.

PR #4010, opened 2019-05-14
fix(FEC-9085): playState is marked as "paused" although playing

shalommeoded:
Added isIPad checker to the interface update, in case `EmbedPlayer.EnableIpadHTMLControls` config is set to false.

### DIFF
--- a/modules/EmbedPlayer/resources/mw.EmbedPlayerNative.js
+++ b/modules/EmbedPlayer/resources/mw.EmbedPlayerNative.js
@@ -1309,9 +1309,15 @@
 			}
 
 			// Update the interface ( if paused )
-			if (!this.ignoreNextNativeEvent && this._propagateEvents && this.paused || (mw.isIpad() && ( mw.getConfig('EmbedPlayer.EnableIpadHTMLControls') === true ))) {
-				this.parent_play();
-			} else {
+			if (mw.isIpad() && ( mw.getConfig('EmbedPlayer.EnableIpadHTMLControls') === true )) {
+				if (!this.ignoreNextNativeEvent && this._propagateEvents && this.paused ) {
+					this.parent_play();
+				}
+			}
+			if (!this.ignoreNextNativeEvent && this._propagateEvents && this.paused) {
+					this.parent_play();
+			}
+			else {
 				// make sure the interface reflects the current play state if not calling parent_play()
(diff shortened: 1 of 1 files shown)


modules/EmbedPlayer/resources/mw.EmbedPlayerNative.js
```diff
@@ -1309,7 +1309,7 @@
 			}
 
 			// Update the interface ( if paused )
-			if (!this.ignoreNextNativeEvent && this._propagateEvents && this.paused && ( mw.getConfig('EmbedPlayer.EnableIpadHTMLControls') === true )) {
+			if (!this.ignoreNextNativeEvent && this._propagateEvents && this.paused || (mw.isIpad() && ( mw.getConfig('EmbedPlayer.EnableIpadHTMLControls') === true ))) {
```
OrenMe: Why OR?

shalommeoded: Didn't wanted to remove the `EnableIpadHTMLControls` condition since I do not know what it does.
So I added the OR condition so we can enter this condition event if the `EnableIpadHTMLControls` is set to `false` and if we are using iPad.

Do you rather I will just separate the iPad conditioning to a different line?

OrenMe: As this condition should only be evaluated when we are on ipad then it's better separate it outside the if condition.
If we are not on iPad then always return true for it.

modules/EmbedPlayer/resources/mw.EmbedPlayerNative.js
```diff
@@ -1309,9 +1309,15 @@
 			}
 
 			// Update the interface ( if paused )
-			if (!this.ignoreNextNativeEvent && this._propagateEvents && this.paused && ( mw.getConfig('EmbedPlayer.EnableIpadHTMLControls') === true )) {
```
OrenMe: Why not just add a ternary expression before:
```javascript
var enableIpadHTMLControls = mw.isIpad() ?
  ( mw.getConfig('EmbedPlayer.EnableIpadHTMLControls') === true ) : true;

if (!this.ignoreNextNativeEvent && 
  this._propagateEvents && this.paused && 
  enableIpadHTMLControls) {
```

PR #4017, opened 2019-05-26
Fev 304 kava UUID

eitanavgil:


modules/KalturaSupport/UiConfResult.php
```diff
@@ -101,11 +101,31 @@ function loadUiConf() {
...
+			$this->addPartnerPublicInfoData();
 		} else {
 			$this->parseUiConfXML( $this->uiConfFile );
 			$this->setupPlayerConfig();
 		}
 	}
+	/**
+	 * Get partner public data - retrieve analyticsPersistentSessionId flag for kAnalony
+	 */
+	public function addPartnerPublicInfoData(){
```
OrenMe: You are adding another blocking http call, are you sure this is what you meant?
Need to consult BE about added latency and about cache-ability, I guess you don't want it...
But if a session is already available in the client side you will still call this service every time, and taking into consideration the fact this is another blocking call again, I think this should be reviewed.

eitanavgil: @OrenMe I consulted BE - they say this is a very simple and quick API call and there is now. I am merging this into my branch (analytics-2.76)  
We need to release this asap so we start collect real data. I will try to apply the local later - we will not merge it to the main branch until we resolve this

PR #4023, opened 2019-06-04
FEV-128 - set interval for updating Q&A timestamp

semarche-kaltura:


kaltura-hooks:
Hi @semarche-kaltura,

modules/QnA/resources/qnaModule.js
```diff
@@ -145,6 +148,13 @@
             },
             destroy: function () {
                 $(this.embedPlayer).unbind(this.bindPostfix);
+                clearInterval(this.intervalId);
+            },
+            initUpdateInterval: function () {
+                // set interval for updating Q&A timestamp
+                this.intervalId = setInterval(() => {
```
eitanavgil: @semarche-kaltura 
mwEmbed does not transpile and we use the code as is, which means that we cannot use ES2016. Lets remove the  fat-arrow implementation

PR #4029, opened 2019-06-24
An 738 bitrate is na n when single flavor

eitanavgil:


modules/KalturaSupport/resources/mw.KWidgetSupport.js
```diff
@@ -858,7 +858,11 @@ mw.KWidgetSupport.prototype = {
 					var flavors = playmanifest.flavors;
 					if ( flavors && flavors.length === 1 ) {
 						srcToPlay = flavors[0].url;
-						deferred.resolve( srcToPlay );
+						if(flavors.length === 1 && flavors[0].bitrate ){
```
OrenMe: You already checked for length===1 two lines above so this is redundant

modules/Hlsjs/resources/hlsjs.js
```diff
@@ -585,6 +592,12 @@
 							'data-assetid': index
 						};
 					});
+
+					// [Oren] should we check this.embedPlayer.isLive() too? any risk for VOD?
```
OrenMe: This is good for both as this will also happen to single bitrate vod entries

modules/KalturaSupport/resources/mw.KWidgetSupport.js
OrenMe: Also, the bitrate here is in kbps IIRC so normalize it already to bps as it comes like this in the manifests

modules/Hlsjs/resources/hlsjs.js
OrenMe: please remove the comment

modules/Hlsjs/resources/hlsjs.js
```diff
@@ -28,7 +28,7 @@
 			/** @type {Number} */
 			mediaErrorRecoveryCounter: 0,
 			playerErrorRecoveryCounter: 0,
-
+			localSingleFlavor : undefined,
```
OrenMe: better use null as undefined is the default for uninitialised members.

PR #4032, opened 2019-07-02
fix(FEV-314): Save button was disabled upon clicking the hint icon

shalommeoded:
Fixed the issue with the Save and Clear button, not appearing when clicking on the "Hint" button

### DIFF
--- a/modules/Quiz/resources/quiz.js
+++ b/modules/Quiz/resources/quiz.js
@@ -557,6 +557,7 @@
                     $(".close-button")
                     .click(function () {
                         interfaceElement.find(".open-question-textarea").val(holdAnswer);
+                        interfaceElement.find("#open-question-clear,#open-question-save").removeAttr("disabled");
                     })
                 });
             });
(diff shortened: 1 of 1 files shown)


shalommeoded:
@eitanavgil 
Fixed that issue

modules/Quiz/resources/quiz.js
```diff
@@ -557,6 +557,10 @@
                     $(".close-button")
                     .click(function () {
                         interfaceElement.find(".open-question-textarea").val(holdAnswer);
+                        interfaceElement.find(".open-question-chars .chars").text(holdAnswer.length);
+                        if (holdAnswer.length !== 0) {
+                            interfaceElement.find("#open-question-clear,#open-question-save").removeAttr("disabled");
```
eitanavgil: @shalommeoded - looks better. 
Have you tested that also when you visit an answer that was already submitted? (like  when you are reviewing the answer and cannot re-answer it)

shalommeoded: Yes and also when you have already answered, the UI will change from `Clear` and `Submit` to
`Change answer`

PR #4034, opened 2019-07-08
fix(SUP-18895): Vertical Space is cutting Thai letters

shalommeoded:
Added additional line spacing for the default Closed Captioning style in order to support Thai Characters.

### DIFF
--- a/modules/KalturaSupport/resources/mw.ClosedCaptions.js
+++ b/modules/KalturaSupport/resources/mw.ClosedCaptions.js
@@ -1015,7 +1015,6 @@
 		getDefaultCaptionCss: function(){
 			var style = {};
 			style["display"] = "inline-block";
-			style["line-height"] = "2";
 			if (this.getConfig('bg')) {
 				style["background-color"] = mw.getHexColor(this.getConfig('bg'));
 			}
--- a/modules/KalturaSupport/resources/mw.ClosedCaptions.js
+++ b/modules/KalturaSupport/resources/mw.ClosedCaptions.js
@@ -1014,7 +1014,8 @@
 		},
 		getDefaultCaptionCss: function(){
 			var style = {};
-			style["display"] = "inline";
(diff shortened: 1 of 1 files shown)


OrenMe:
@yairans PTAL

modules/KalturaSupport/resources/mw.ClosedCaptions.js
```diff
@@ -1014,7 +1014,8 @@
 		},
 		getDefaultCaptionCss: function(){
 			var style = {};
-			style["display"] = "inline";
+			style["display"] = "inline-block";
+			style["line-height"] = "2";
```
yairans: @shalommeoded from my test [here](http://www.kaltura.com/index.php/extwidget/preview/partner_id/2290881/uiconf_id/39595521/entry_id/1_epi34r91/embed/dynamic?debugKalturaPlayer) `dispaly:inline-block` is enough. no?

shalommeoded: @yairans  in some cases, Thai letters can have up to 3-4 additional symbols on top of the character, therefore, I have added a bit more line space to the background.

yairans: again, as is sow `display:inline-block` solves the issue. did you see else?

shalommeoded: Yes, in Thai letters, the background will cut in some cases the Thai symbols. If you can see that `display:inline-block` is enough I can alter the fix.

yairans: This is what I see but better you verify by yourself

shalommeoded: @yairans 
Just checked, you are right, the difference does not justify to enhance the line-height.

PR #4037, opened 2019-07-09
Webc 1206

eitanavgil:


modules/EmbedPlayer/resources/mw.EmbedPlayerNative.js
```diff
@@ -1611,11 +1611,17 @@
 			var _this = this;
 			metadataTrack.addEventListener("cuechange", function (evt) {
 				try {
-					if(!evt.currentTarget.activeCues.length){
-						return;
+					var id3Tag;
+					if ( mw.isEdge() ){
```
OrenMe: we can at least use the `actievCues` for both or this is also an issue?
Also, can you please add bug description.

PR #4038, opened 2019-07-09
FEV-320 - start playing on a question-cue-point click

semarche-kaltura:
on CP click it seeked 0.5 sec before the question cue-point marker, so in case when video is paused - it seeked 0.5 sec before the CP and stayed paused.
The changes makes sure that when we seek on a question-cue-point it always keeps playing and not paused.


kaltura-hooks:
Hi @semarche-kaltura,

PR #4041, opened 2019-07-16
[DO NOT MERGE]  WEBC-1206 live video appears before the slide does.

eitanavgil:


modules/KalturaSupport/components/dualScreen/imagePlayer.js
```diff
@@ -104,6 +105,26 @@
...
+				// if we stored a preview thumb-cuepoint - restore and clear it
+				_this.sync(_this.latestPreviewCuepoint);
+					_this.latestPreviewCuepoint = null;
+				}
+			});
+
+			// listen to push cuepoint API
+			this.bind('KalturaSupport_PushCuePointsReceived' , function(event,cuepoint){
+				if( !cuepoint.length && cuepoint && cuepoint.tags
+					&& _this.embedPlayer.isLive()
```
noam-arad: isn't this supposed to fix the problem when the player is not live?

PR #4045, opened 2019-07-23
feat(FEC-9034): don't filter same language audio track

OrenMe:
remove filtering from hls as it was never an issue there
changed the way we filter on shaka to use built in methods that only give audio representations languages and not adaptations.

### DIFF
--- a/modules/Dash/resources/dash.js
+++ b/modules/Dash/resources/dash.js
@@ -475,7 +475,7 @@
 				if ( embedPlayer.isLive() && embedPlayer.isDVR() ) {
 					// embedPlayer.duration is irrelevant for dash live, we have to calculate the playHeadPercent via player.seekRange().end instead.
 					if ( embedPlayer.currentTime >= 0 && embedPlayer.duration && !embedPlayer.userSlide && !embedPlayer.seeking ) {
-						var delta = player.seekRange().end - embedPlayer.currentTime;
+						var delta = this._shaka.seekRange().end - embedPlayer.currentTime;
 						var playHeadPercent = ( embedPlayer.duration - delta ) / embedPlayer.duration;
 						embedPlayer.updatePlayHead(playHeadPercent);
 						//update liveEdgeOffset
--- a/modules/Dash/resources/dash.js
+++ b/modules/Dash/resources/dash.js
@@ -178,9 +178,9 @@
 			//Reinstall the polyfills to make sure they weren't ran over by others(VTT.js runs over VTTCue polyfill)
 			shaka.polyfill.installAll();
 			// Create a Player instance.
-			var player = new shaka.Player(this.getPlayer().getPlayerElement());
+			var player = this._shaka = new shaka.Player(this.getPlayer().getPlayerElement());
 
-				player.configure(this.getShakaConfig());
+				this._shaka.configure(this.getShakaConfig());
 
 				// Attach player to the window to make it easy to access in the JS console.
 				window.player = player;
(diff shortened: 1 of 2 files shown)


OrenMe:
need to apply same code we use in V7 to filter out same dash AdaptationSet representation for audio so we won;t show same language with different bitrates as selection in menu.

modules/Hlsjs/resources/hlsjs.js
```diff
@@ -331,20 +331,19 @@
...
-								'kind': 'audioTrack',
-								'language': audioTrack.lang,
-								'srclang': audioTrack.lang,
-								'label': audioTrack.name,
-								'title': audioTrack.name,
-								'id': audioTrack.id,
-								'index': audioTrackData.languages.length
-							});
-						}
+						audioTrackData.languages.push(createAudioTrack(audioTrack));
```
yairans: same comment as in dash

modules/Hlsjs/resources/hlsjs.js
```diff
@@ -331,20 +331,19 @@
...
+							'srclang': audioTrack.lang,
+							'label': audioTrack.name,
+							'title': audioTrack.name,
+							'id': audioTrack.id,
+							'index': audioTrackData.languages.length
+						};
+					};
 					$.each(audioTracks, function (index, audioTrack) {
-						if (audioTrackLangs[audioTrack.lang] === undefined) {
-							audioTrackLangs[audioTrack.lang] = 1;
```
yairans: What's about duplicate audios?

modules/Dash/resources/dash.js
```diff
@@ -178,9 +178,9 @@
 			//Reinstall the polyfills to make sure they weren't ran over by others(VTT.js runs over VTTCue polyfill)
 			shaka.polyfill.installAll();
 			// Create a Player instance.
-			var player = new shaka.Player(this.getPlayer().getPlayerElement());
+			var player = this._shaka = new shaka.Player(this.getPlayer().getPlayerElement());
```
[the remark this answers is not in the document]
OrenMe: player is still exposed on window and we can't know if anyone is accessing it like this - don't want to break it

modules/Hlsjs/resources/hlsjs.js
OrenMe: As I wrote, it was never an issue in hls, just dash that we had multi bitrate(representation) per one audio adaptation, this doesn't exist at all in hls.

modules/Dash/resources/dash.js
```diff
@@ -284,20 +287,19 @@
...
-								'kind': 'audioTrack',
-								'language': audioTrack.language,
-								'srclang': audioTrack.language,
-								'label': audioTrack.label || audioTrack.language,
-								'title': audioTrack.language,
-								'id': audioTrack.id,
-								'index': audioTrackData.languages.length
-							});
-						}
+						audioTrackData.languages.push(createAudioTrack(audioTrack));
```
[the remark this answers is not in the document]
OrenMe: because we used to filter out some indexes we couldn't pass them, now that we don't this is totally valid. done.

modules/Hlsjs/resources/hlsjs.js
OrenMe: done.

modules/Dash/resources/dash.js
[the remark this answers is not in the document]
yairans: sure but you can do
`window.player = this._shaka`
my question was about the `var player`

modules/Hlsjs/resources/hlsjs.js
yairans: As I remember we had 2 'undefined' audio labels. this scenario cannot be with hls?

modules/Dash/resources/dash.js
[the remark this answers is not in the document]
OrenMe: No change there, we are putting it in a `var` and then doing `window.player = player`

PR #4046, opened 2019-07-24
fix(SUP-18987): issue with aspect in 360

shalommeoded:
Issue with aspect calculation when 360 video is on a playlist, the aspect is miscalculated (includes the hight or width of the playlist) and results in a malformed video display.

The calculation was if the original calculation was less than 1.4, we will set it to 1.4

### DIFF
--- a/modules/KalturaSupport/components/360/resources/video360.js
+++ b/modules/KalturaSupport/components/360/resources/video360.js
@@ -15,7 +15,7 @@
             mobileVibrationValue: (mw.isAndroidNativeBrowser() ? 1 : 0.02),
 			cameraOptions: {
 				fov: 75,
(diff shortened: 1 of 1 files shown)


modules/KalturaSupport/components/360/resources/video360.js
```diff
@@ -1,8 +1,9 @@
 (function (mw, $, THREE) {
 	"use strict";
 
-	mw.PluginManager.add('video360', mw.KBaseComponent.extend({
+	var DEFAULT_ASPECT = Math.max(640/360);
```
yairans:
```suggestion
	var DEFAULT_ASPECT = 640/360;
```

shalommeoded: Changed

PR #4001, opened 2019-04-14
fix(SUP-17808): No playback on progressive when changing stream quality  

shalommeoded:
Fuflied the required play promise when on pause, when the player is seeking.

PR #4052, opened 2019-07-29
fix(FEC-9241): entry is replaying on IOS after post-roll

shalommeoded:
We trigged the vid.play() on iOS and caused a replay after the postroll.

### DIFF
--- a/modules/EmbedPlayer/resources/mw.EmbedPlayerNative.js
+++ b/modules/EmbedPlayer/resources/mw.EmbedPlayerNative.js
@@ -789,9 +789,6 @@
 							vid.play();
 						} else {
 							_this.removeBlackScreen();
-							if (!mw.isIOS()) {
-								vid.play();
-							}
 						}
 					});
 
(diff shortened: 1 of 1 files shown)


modules/EmbedPlayer/resources/mw.EmbedPlayerNative.js
```diff
@@ -789,7 +789,9 @@
 							vid.play();
 						} else {
 							_this.removeBlackScreen();
-							vid.play();
+							if (!mw.isIOS()) {
+								vid.play();
```
OrenMe: From the comment above it actually sounds like we need this play event and something else should pause the player.

shalommeoded: I actually have added this play event in here:
https://github.com/kaltura/mwEmbed/pull/4001/files

OrenMe: So you should have probably added this original play call with a condition or in another place

PR #4054, opened 2019-07-29
fix(FEC-9244): Issue with Skippable ads on IOS

shalommeoded:
Exposed the `disableCustomPlaybackForIOS10Plus` and set in to true only when we are in playsinline

### DIFF
--- a/modules/DoubleClick/resources/mw.DoubleClick.js
+++ b/modules/DoubleClick/resources/mw.DoubleClick.js
@@ -224,6 +224,12 @@
 	            google.ima.settings.setVpaidMode(vpaidMode);
             };
 
+            var getDisableCustomPlaybackForIOS10Plus = function() {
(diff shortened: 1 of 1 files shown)


modules/DoubleClick/resources/mw.DoubleClick.js
```diff
@@ -224,6 +224,12 @@
 	            google.ima.settings.setVpaidMode(vpaidMode);
             };
 
+            var getDisableCustomPlaybackForIOS10Plus = function() {
+                var _disableCustomPlaybackForIOS10Plus =  !mw.getConfig("EmbedPlayer.WebKitPlaysInline") ? false
```
OrenMe: why negate then? and also what is `google.ima.ImaSdkSettings.disableCustomPlaybackForIOS10Plus = true;`? IMA ImaSdkSettings doesn't have `disableCustomPlaybackForIOS10Plus` property.
Please revise your code and check, suggest you check playkit-js-ima for reference.

PR #4060, opened 2019-09-01
trigger report to Analytics of the default playback rate speed when the player first loads

amirch1:
FYI @itaykinnrot

### DIFF
--- a/modules/KalturaSupport/resources/uiConfComponents/playbackRateSelector.js
+++ b/modules/KalturaSupport/resources/uiConfComponents/playbackRateSelector.js
@@ -51,6 +51,7 @@
 				this.speedSet[i] = Math.abs( this.speedSet[i] );
 			}
 			this.addBindings();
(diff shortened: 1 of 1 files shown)


modules/KalturaSupport/resources/uiConfComponents/playbackRateSelector.js
```diff
@@ -51,6 +51,7 @@
 				this.speedSet[i] = Math.abs( this.speedSet[i] );
 			}
 			this.addBindings();
+            this.getPlayer().triggerHelper( 'updatedPlaybackRate', this.currentSpeed); // trigger report to Analytics
```
OrenMe: `updatedPlaybackRate` event means that the speed was changed from the plugin, which a user action, so not sure if adding the event like this is the correct way to go.
Probably need to add API to get rate on the base player and query it on analytics plugin CTOR

PR #4062, opened 2019-09-04
fix(SUP-18477): save input if user is clicking of the why button a second time

shalommeoded:
@einatr 
I've changed the `focus` from the question to the open-area element and that will save the question even if the user clicked the question a second time without changing the answer.


eitanavgil:
@shalommeoded  - nice solution 
@nirbe - i want to test this in person (like to run this) so we dont have regressions we missed. I need the time for it ±1-2h

PR #4065, opened 2019-09-15
FEV-221

eitanavgil:
on selecting a chapter with the keyboard prevent closing the menu


shalommeoded:
@eitanavgil 
Looks good.

PR #4066, opened 2019-09-16
fix(SUP-19385): Question is skipped when selecting an answer and using the Skip button

shalommeoded:
Issue when a user is selecting an answer and choosing to skip to the next question at the same time, the next Question is skipped.

### DIFF
--- a/modules/Quiz/resources/quiz.js
+++ b/modules/Quiz/resources/quiz.js
@@ -936,6 +936,8 @@
             });
             _this.KIVQModule.submitAnswer(questionNr,_this.selectedAnswer);
             _this.selectedAnswer = null;
+            $(".cp-navigation").css("display", "none");
+            $(".ftr-container").css("height", "40px");
             setTimeout(function(){_this.KIVQModule.checkIfDone(questionNr)},_this.postAnswerTimer);
         },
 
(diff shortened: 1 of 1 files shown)


modules/Quiz/resources/quiz.js
```diff
@@ -936,6 +936,8 @@
             });
             _this.KIVQModule.submitAnswer(questionNr,_this.selectedAnswer);
             _this.selectedAnswer = null;
+            $(".cp-navigation").css("display", "none");
+            $(".ftr-container").css("height", "40px");
```
eitanavgil: @shalommeoded - we need to make sure this code does not interfere with all 4 question types 
If you have checked that - I can review

shalommeoded: Checked:
	
         1. On Multiple it works, just fine.
	2. On True and False as well, since it uses the same logic.
	3. Reflection point, not relevant.
	4. Same for open Q

shalommeoded:
@eitanavgil 
Per our discussion with Ido, I have altered the behavior and the skip arrows will still get displayed and now will just get grayed out.

1.  Added the disabled class to the arrows.
2. Added pointer-events: none to as well (the disable class will just alter the appearance and not the functionality).

PR #4068, opened 2019-09-22
fix(SUP-19662): "Off air" string is showing on non-Live entries

shalommeoded:
@OrenMe 
It seems that the setLiveStatusButtonDefaultState() is bindded when onChange is getting sent 
(For example: https://github.com/kaltura/mwEmbed/blob/e667d18f51107074fbcdef0c6e5b5e7b543d242a/modules/KalturaSupport/components/related/related.js#L420)

I've added a condition to only apply to LiveStream type entries.

### DIFF
--- a/modules/KalturaSupport/components/live/liveStatus.js
+++ b/modules/KalturaSupport/components/live/liveStatus.js
@@ -37,8 +37,10 @@
 			this.bind( 'onChangeMedia', function() {
                 _this.removeBindings();
 				//Reset UI state on change media and liveStatus button
-				_this.setLiveStatusButtonDefaultState();
-				_this.getBtn().show();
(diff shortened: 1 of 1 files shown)


modules/KalturaSupport/components/live/liveStatus.js
```diff
@@ -37,8 +37,10 @@
 			this.bind( 'onChangeMedia', function() {
                 _this.removeBindings();
 				//Reset UI state on change media and liveStatus button
-				_this.setLiveStatusButtonDefaultState();
-				_this.getBtn().show();
+				if( _this.getPlayer().isLive() ) {
```
OrenMe: don't you want remove it if entry is not live in the onChangeMedia done event?

PR #4069, opened 2019-09-22
FEV-381 - CVAA preview-screen accessibility

eitanavgil:


modules/KalturaSupport/components/cvaa/cvaa.js
```diff
@@ -228,6 +229,11 @@
                         _this.initPreviewUpdate(_this.getCurrentPreset(_this.cvaaSettings.currentPreset));
                         _this.saveCvaaSettings();
                     }
+
+                    if(event.which === 32 || event.which === 13){
```
semarche-kaltura: same logic on line 226, we could move this condition to separate method

eitanavgil: @semarche-kaltura 
not exactly 
line 226 covers click, enter and space, while here - I want to close the screen only if user uses the keyboard. If he used a mouse-click I do not want to close the screen.

PR #4072, opened 2019-10-02
PLAT-9400 split flavorAsset multi-request

semarche-kaltura:
Split flavorAsset API call into 2 different calls:
1) get child entries
2) in case if child entries exist - get flavorAsset for child entries

### DIFF
--- a/modules/KalturaSupport/components/streamSelectorUtils.js
+++ b/modules/KalturaSupport/components/streamSelectorUtils.js
@@ -74,30 +74,41 @@
         },
         getStreams: function () {
             var _this = this;
-            var requestObject = [];
-            requestObject.push({
-                'service': 'baseEntry',
-                'action': 'list',
-                'filter:objectType': 'KalturaBaseEntryFilter',
-                'filter:typeEqual': 1,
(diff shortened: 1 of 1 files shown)


modules/KalturaSupport/components/streamSelectorUtils.js
```diff
@@ -74,30 +74,41 @@
...
+            _this.getKalturaClient().doRequest(requestObject, function (data) {
                 // Validate result
-                if (data && _this.isValidResult(data[0] && data[0].totalCount > 0)) {
-                    _this.createStreamList(data);
+                if (data && _this.isValidResult(data && data.totalCount > 0)) {
+                    responceArray.push(data)
+                    var i = 0;
+                    var maxNumOfStream = _this.getConfig("maxNumOfStream");
+                    var flavorAssetRequestObject = [];
+                    for (i; i < data.totalCount && i < maxNumOfStream; i++) {
```
eitanavgil: @semarche-kaltura 
consider use 
maxNumOfStream = math.Max(maxNumOfStream,data.totalCount) and then in the for-loop use a simpler expression

semarche-kaltura: done

PR #4075, opened 2019-10-03
fix(FEC-9386): Unmuted button is not working on iPad iOS13

shalommeoded:
@OrenMe 

I made the the iPad on iOS 13 get recognized as a mobile device, see if that is ok.

### DIFF
--- a/modules/MwEmbedSupport/mediawiki/mediawiki.client.js
+++ b/modules/MwEmbedSupport/mediawiki/mediawiki.client.js
@@ -8,7 +8,7 @@
 	var userAgent = navigator.userAgent;
 
 	mw.isMobileDevice = function () {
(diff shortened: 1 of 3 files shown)


modules/MwEmbedSupport/mediawiki/mediawiki.client.js
```diff
@@ -82,7 +82,10 @@
 	mw.isIpad3 = function () {
 		return  /OS 3_/.test(userAgent) && mw.isIpad();
 	};
-	
+	mw.isIpadOnIos13 = function() {
+		return mw.isSafari() && mw.isTouchDevice();
```
OrenMe: this will be true for iOS as well, no? and it will be correct also for 14, 15 an onwards so isIpadOS is a better name

PR #4076, opened 2019-10-07
Fev 370 playlist accessibility

semarche-kaltura:
- able to navigate with tabs over the list items 
- if item is in focused - enter will activate it  
- access the playlist buttons next / prev button with TAB
- if playlist name is link - added role="link" 
- after clicking (by space or by enter) focus moved to the little play button 
- aria-labels for playlist items

Screen Shot 2019-10-07 at 12 33 52 PM


jbrule:
We had the Kaltura player audited for ADA level AA and the playlist was flagged as failing to be accessible. I came here to see if there was anything in the pipeline regarding the issue and came upon this. Is it safe to assume this addresses that need?

Thank you,
Josh

PR #4095, opened 2019-11-11
fix(SUP-20146): Accessibility changes to the displayed question

shalommeoded:
Per Ido, we want to alter the Accessibility method of the display question from Aria-label to aria-labelledby.

### DIFF
--- a/modules/Quiz/resources/quiz.js
+++ b/modules/Quiz/resources/quiz.js
@@ -447,7 +447,7 @@
             }
             var displayQuestion = $(".display-question");
             //Search for links and links patterns ( [text|http://exampl.com] ) and add a real link to a new tab
-            displayQuestion.text(cPo.question).attr({'tabindex': 5,"aria-labelledby":cPo.question}).focus();
+            displayQuestion.text(cPo.question).attr({'tabindex': 5,"aria-labelledby":text(cPo.question)}).focus();
             var questionText = this.wrapLinksWithTags(displayQuestion.html());
             displayQuestion.html(questionText);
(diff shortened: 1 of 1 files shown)


modules/Quiz/resources/quiz.js
```diff
@@ -447,7 +447,7 @@
             }
             var displayQuestion = $(".display-question");
             //Search for links and links patterns ( [text|http://exampl.com] ) and add a real link to a new tab
-            displayQuestion.text(cPo.question).attr({'tabindex': 5,"aria-lable":cPo.question}).focus();
+            displayQuestion.text(cPo.question).attr({'tabindex': 5,"aria-labelledby": cPo.question}).focus();
```
eitanavgil: @shalommeoded 
labelledby means that the label for accessibility will be taken from another HTML element. 
For example 
```
<div id="myBillingId">Billing</div>

<div>
    <div id="myNameId">Name</div>
    <input type="text" aria-labelledby="myBillingId myNameId"/>
</div>
<div>
    <div id="myAddressId">Address</div>
    <input type="text" aria-labelledby="myBillingId myAddressId"/>
</div>
```

It does not make sense that only the attribute was changed from aria-label to aria-labelledby without any reference to the content of it. This will break the screen readers. 

I suggest to understand better the difference between aria-label and aria-labelledby, to query the structure of the HTML and see if this is possible as aria-labelledby is expected to have an id of another element.

shalommeoded: Yeah, I have researched the matter and it seems that the usage of aria-label was correct however, it was misspelled.

PR #4096, opened 2019-11-11
fix(FEC-9490): ChromeCast toggle button does not appear

shalommeoded:
In some cases the plugin will render when the player is not ready, and will not display

### DIFF
--- a/modules/Chromecast/resources/sender/chromecast.js
+++ b/modules/Chromecast/resources/sender/chromecast.js
@@ -35,23 +35,26 @@
          * Setup the Chromecast plugin - loads Google Cast Chrome sender SDK and bind his handler.
          */
         setup: function () {
-            if ( mw.getConfig( 'EmbedPlayer.IsFriendlyIframe' ) ) {
-                try {
-                    top[ '__onGCastApiAvailable' ] = this.toggleCastButton.bind( this );
-                    kWidget.appendScriptUrl( this.CAST_SENDER_V3_URL, null, top.document );
-                    this.isInIframeApi = false;
-                } catch ( e ) {
-                    window[ '__onGCastApiAvailable' ] = this.toggleCastButton.bind( this );
-                    // PSVAMB-4560
-                    // For some reason, Chrome appends script to a player's iframe parent document unless specified explicitly
-                    kWidget.appendScriptUrl( this.CAST_SENDER_V3_URL, null, window.document );
-                    this.isInIframeApi = true;
+            var _this = this;
+            this.bind('playerReady', function() {
+                if (mw.getConfig('EmbedPlayer.IsFriendlyIframe')) {
+                    try {
+                        top['__onGCastApiAvailable'] = _this.toggleCastButton.bind(_this);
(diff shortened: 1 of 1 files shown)


OrenMe:
@shalommeoded i don’t u defat and why the playerReady has any relation to the issue, can you please explain?

shalommeoded:
@OrenMe 
In some cases the Chromecast plugin is trying to render when the player is not ready, and causing the plugin not to show.

OrenMe:
@shalommeoded can you explain the flow that is not working?
You just delayed the loading of the 3rd party cast lib, which is async anyway, and the render is actually very deterministic for kbaseconponent.
Waiting for player ready feels a bit like setTimeout magic.

shalommeoded:
@OrenMe Sure,
From what I was able to debug, I saw that in some instances when we enter [here](https://github.com/kaltura/mwEmbed/blob/62ca0b92fe08647b8661f751771c272abbaf1355/modules/Chromecast/resources/sender/chromecast.js#L73-L75), and enter the show function **before** the player is ready, the plugin will not show sometimes, so we may do need some setTimeout here.

We can maybe isolate the toggle binding to addBindings function if it's a better practice.

modules/Chromecast/resources/sender/chromecast.js
```diff
@@ -69,10 +69,16 @@
          * @param reason
          */
         toggleCastButton: function ( isAvailable, reason ) {
+            var _this = this;
             this.log( "toggleCastButton: isAvailable=" + isAvailable + ", reason=" + reason );
             if ( isAvailable ) {
                 this.initializeCastApi();
                 this.show();
+                if (this.embedPlayer.getControlBarContainer().length < 1) {
+                    this.bind('onShowInterfaceComponents', function () {
```
OrenMe: this is an event which is a side effect of the live core, don't see it used anywhere else.
Why not check the `embedPlayer.playerReadyFlag` and call show or wait to playerReady event according to it.

BTW, same goes to hide I guess.

shalommeoded: Done.
I used the onShowInterfaceComponents cause I saw the same event [here](https://github.com/kaltura/mwEmbed/blob/17765fee3badb1ea61e41542ec320d14d2339231/modules/KalturaSupport/resources/mw.KBaseComponent.js#L96-L100) and though it will be a better to use.

modules/Chromecast/resources/sender/chromecast.js
```diff
@@ -69,12 +69,13 @@
          * @param reason
          */
         toggleCastButton: function ( isAvailable, reason ) {
+            var _this = this;
             this.log( "toggleCastButton: isAvailable=" + isAvailable + ", reason=" + reason );
             if ( isAvailable ) {
                 this.initializeCastApi();
-                this.show();
+                this.embedPlayer.playerReadyFlag ? this.show() : this.bind('playerReady', function (){_this.show()});
```
OrenMe: I would prefer to keep it DRY, so create an inline function called loadWhenReady and call it with the callback function as a prameter.

shalommeoded: Done

PR #4098, opened 2019-11-13
fix(FEV-401): Chapters are overlapping Vertical Volume Control

shalommeoded:
The change was done with the Z-index prioritization.


eitanavgil:
@shalommeoded 
I am OK with merging this - but we have no idea what side-effect will happen with this change and as a developer - I cannot predict them. 
We will need to make sure QA understand this and cover as many use-cases as possible.

PR #4106, opened 2019-12-18
Fev 398 quiz loading animation

semarche-kaltura:


modules/Quiz/resources/quiz.js
```diff
@@ -502,6 +504,19 @@
                     _this.showAnswered(cPo, questionNr);
                 }
             }
+            else if (cPo.submitRequested) {
```
eitanavgil: long if-else logic. Consider adding a short comment of the case so it will be more readable

modules/Quiz/resources/quiz.js
```diff
@@ -856,16 +871,24 @@
...
-                    $('#' + key).parent().addClass("wide single-answer-box-bk-apply disable");
-                    $('#' + key).removeClass('single-answer-box-txt')
-                        .addClass(function(){
-                            // transform answer element so it appears as an already-answered on with "applied" element
-                            $(this).addClass('single-answer-box-txt-wide ')
-                                .after($('<div></div>') // adding continue/applied div as button
-                                    .addClass("single-answer-box-apply qApplied disable").attr('aria-disabled', true).attr('role', 'button')
-                                    .text(gM('mwe-quiz-selected'))
-                            );
+                    if (cPo.submitRequested) {
```
eitanavgil: Do we set submitRequested to false at any point (other than initial) ? 
Do we need to?

modules/Quiz/resources/quiz.js
```diff
@@ -937,12 +971,18 @@
                     .text(gM('mwe-quiz-selected'))
                     .addClass('qApplied').fadeIn(100).attr('aria-disabled', true);
             });
-            _this.KIVQModule.submitAnswer(questionNr,_this.selectedAnswer);
-            _this.selectedAnswer = null;
-            $(".cp-navigation").css("pointer-events", "none");
-            $(".ftr-right").css("pointer-events", "none");
-            $(".cp-navigation-btn").addClass("disabled");
-            setTimeout(function(){_this.KIVQModule.checkIfDone(questionNr)},_this.postAnswerTimer);
+            _this.KIVQModule.submitAnswer(questionNr,_this.selectedAnswer).then(
```
eitanavgil: Are we doing this on all questions types? Should we not do this only on reflection-point questions?

modules/Quiz/resources/IVQModule.js
```diff
@@ -360,22 +362,24 @@
...
-                }
-                else{
-                    isAnswered = false;
-                    $.cpObject.cpArray[questionNr].isAnswerd = true;
-                }
-                _this.KIVQApi.addAnswer(isAnswered,_this.i2q(selectedAnswer),_this.kQuizUserEntryId,questionNr,function(data){
+                $.cpObject.cpArray[questionNr].submitRequested = true;
+                var submitCallTimeout = setTimeout(function(){
+                    defer.reject();
+                }, _this.quizPlugin.getConfig("apiRequestTimeout") || 5000)
```
eitanavgil: better explicitly add a new configuration and set the default there instead of checking this with every time we submit answer

modules/Quiz/resources/IVQModule.js
```diff
@@ -360,22 +362,24 @@
...
+                _this.KIVQApi.addAnswer($.cpObject.cpArray[questionNr].isAnswerd,_this.i2q(selectedAnswer),_this.kQuizUserEntryId,questionNr,function(data){
                     if (!_this.checkApiResponse('Add question err -->',data)){
-                        return false;
-                    }else {
+                        defer.reject();
+                    } else {
+                        clearTimeout(submitCallTimeout);
+                        $.cpObject.cpArray[questionNr].isAnswerd = true;
                         $.cpObject.cpArray[questionNr].answerCpId = data.id;
+                        defer.resolve();
```
eitanavgil: @semarche-kaltura  we still need to think if we can handle success callback after timeout. very edge case where the BE responses really slow but still possible. 
I will try to think with Eran on this.

modules/Quiz/resources/quiz.js
semarche-kaltura: I think we shouldn't care about it, because "submitRequested" flag creates and works only for the current quiz session. In case if API respond with delay - question get "isAnswerd" flag then "submitRequested" flag become unnecessary.

modules/Quiz/resources/quiz.js
semarche-kaltura: logic changed according to new requirements about reflection point

modules/Quiz/resources/IVQModule.js
```diff
@@ -360,22 +363,24 @@
...
+                $.cpObject.cpArray[questionNr].submitRequested = true;
+                var submitCallTimeout = setTimeout(function(){
+                    defer.reject();
+                }, _this.apiRequestTimeout);
+                _this.KIVQApi.addAnswer($.cpObject.cpArray[questionNr].isAnswerd,_this.i2q(selectedAnswer),_this.kQuizUserEntryId,questionNr,function(data){
+                    clearTimeout(submitCallTimeout);
                     if (!_this.checkApiResponse('Add question err -->',data)){
-                        return false;
-                    }else {
+                        defer.reject();
```
eitanavgil: @semarche-kaltura 
If we fail - do we still want to hold the "submitRequested" flag as true? Will that not make bugs on other logic?

modules/Quiz/resources/IVQModule.js
eitanavgil: We have a BE request to return an BE error when we try to submit the same question twice. Currently it is allowed 
https://kaltura.atlassian.net/browse/PLAT-10458

modules/Quiz/resources/css/quiz.css
```diff
@@ -309,6 +313,9 @@ button.qContinue, button.qApplied {
...
     background-size: 100% 100%;
 }
+.single-answer-box-bk.disable{
```
eitanavgil: @semarche-kaltura 
What is that for?

PR #4113, opened 2020-01-14
fix(SUP-20341): analytics showing incorrect media type on changeMedia

shalommeoded:
Added an additional flag _mediaChange that will be a condition when sendingAnalytics and will set to true onChangeMedia and will reset to false on resetPlayerflags.

### DIFF
--- a/modules/KalturaSupport/resources/mw.kAnalony.js
+++ b/modules/KalturaSupport/resources/mw.kAnalony.js
@@ -577,143 +577,144 @@
         },
 
 		sendAnalytics : function(eventType, additionalData){
-			if (!this._mediaChange) {
-				//Don't send analytics if entry or partner id are missing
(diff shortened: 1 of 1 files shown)


modules/KalturaSupport/resources/mw.kAnalony.js
```diff
@@ -574,141 +577,143 @@
...
-                'applicationName' : 'application',
-                'userId' : 'userId'
-            };
-            // support legacy ( deprecated ) top level config
-            for( var fvKey in flashVarEvents){
-                if( this.embedPlayer.getKalturaConfig( '', fvKey ) ){
-                    statsEvent[ flashVarEvents[ fvKey ] ] = this.embedPlayer.getKalturaConfig('', fvKey );
-                }
-            }
+			if (!this._mediaChange) {
```
OrenMe: Let's just add 
```javascript
if (this._mediaChange) {
  return;
}
```

PR #4121, opened 2020-02-04
feat(FEV-471): add parameter manifestDownloadTime

eitanavgil:


modules/Hlsjs/resources/hlsjs.js
```diff
@@ -326,6 +326,7 @@
 			            textTracks[i].mode = "disabled";
 		            }
 	            }
+	            this.getPlayer().triggerHelper('hlsManifestLoadedWithStats',data);
```
OrenMe: would prefer we do it generically and fire a manifestDownaloadStats event and do the data parsing in the context of hls.
This way we can also add it to dash(with Shaka) easily.
The event will carry a normalised payload that you can parse safely in the kava analytics plugin.

PR #4135, opened 2020-03-05
fix(SUP-21004): Chrome-cast is unable to stream via Android

shalommeoded:
Same issue that we had on V7, we call resume instead of init on Android


modules/Chromecast/resources/sender/mw.EmbedPlayerChromecast.js
```diff
@@ -152,7 +152,7 @@
             this.castContext = cast.framework.CastContext.getInstance();
             this.castSession = this.castContext.getCurrentSession();
             this.receiverName = this.getReceiverName();
-            if ( this.castSession.getSessionState() === cast.framework.SessionState.SESSION_RESUMED ) {
+            if ( this.castSession.getSessionState() === cast.framework.SessionState.SESSION_RESUMED && !mw.isAndroid()) {
```
yairans: add check for chrome version >= 73

PR #4137, opened 2020-03-08
feat(FEV-472): support forward buffer health

eitanavgil:
THIS PR covers 2 different features (they both have common functions): 
FEV-474-targetBuffer-support
FEV-472-support-forwardBufferHealth

### DIFF
--- a/modules/EmbedPlayer/resources/mw.EmbedPlayer.js
+++ b/modules/EmbedPlayer/resources/mw.EmbedPlayer.js
@@ -3417,6 +3417,14 @@
 			//adaptive bitrate
 			return this.currentBitrate;
 		},
+		
+		getTargetBuffer: function(){
+			var targetBuffer = null;
+			if(this.plugins.hlsjs && this.plugins.hlsjs.targetBuffer()){
+				targetBuffer = this.plugins.hlsjs.targetBuffer();
+			}
+			return targetBuffer;
+		},
 
 		/*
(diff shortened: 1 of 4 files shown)


modules/KalturaSupport/resources/mw.kAnalony.js
```diff
@@ -540,8 +540,55 @@
...
+			return forwardBufferHealth;
+		},
+		
 
+		
+		// samples the hls at realtime 
+		getTargetBuffer : function(){
+			var targetBuffer = null;
+			if(this.embedPlayer.plugins.hlsjs){
+				targetBuffer = this.embedPlayer.plugins.hlsjs.targetBuffer();
```
OrenMe: to avoid this coupling create the interface methods on the embedPlayer and return correct default values there and get the plugin that implements the methods to return correct values according to its implementation

modules/EmbedPlayer/resources/mw.EmbedPlayer.js
```diff
@@ -3417,6 +3417,14 @@
 			//adaptive bitrate
 			return this.currentBitrate;
 		},
+		// get the target buffer from player, supporting only hls.js at this phase. 		
+		getTargetBuffer: function(){
+			var targetBuffer = null;
+			if(this.plugins.hlsjs){
```
OrenMe: See how we implement this in hlsjs, again, the core shouldn't know about the plugin, it should be the other way around.

PR #4142, opened 2020-03-29
fix(KMS-19979): Saves answer after edit for open-question

shalommeoded:
1. I have made it so now we can alter from the hint and question more than once with the setTime on the focus not returning undefined.
2. The original bug when the FE is saving the answer even though we didn't click save.


shalommeoded:
I'll create a new Branch.

PR #4146, opened 2020-04-01
fix(KMS-20622): disable the retake button after first click

shalommeoded:


modules/Quiz/resources/quiz.js
```diff
@@ -294,6 +294,7 @@
 
         retake : function(){
             var _this = this;
+            $(".retake-btn").css("pointer-events", "none");
```
eitanavgil, line 297: @shalommeoded  why CSS and not off() ?

shalommeoded: Changed,
I just used the same method that was used in similar cases.
But they work pretty much the same while testing.

PR #4152, opened 2020-04-28
feat(FEC-10005): investigate flow when isLive XHR call fails

yairans:
Do not move to online (load/restarting hlsjs) unless the player reaches to offline mode (means it has nothing to play)

### DIFF
--- a/modules/EmbedPlayer/resources/mw.EmbedPlayerNative.js
+++ b/modules/EmbedPlayer/resources/mw.EmbedPlayerNative.js
@@ -180,8 +180,11 @@
             this.bindHelper('switchAudioTrack' + this.bindPostfix, function (e, data) {
                 _this.switchAudioTrack(data.index);
             });
-            this.bindHelper('liveOnline' + this.bindPostfix, function () {
-                _this.load();
+            this.bindHelper('liveOffline' + this.bindPostfix, function () {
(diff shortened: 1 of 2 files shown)


modules/EmbedPlayer/resources/mw.EmbedPlayerNative.js
```diff
@@ -180,8 +180,11 @@
             this.bindHelper('switchAudioTrack' + this.bindPostfix, function (e, data) {
                 _this.switchAudioTrack(data.index);
             });
-            this.bindHelper('liveOnline' + this.bindPostfix, function () {
-                _this.load();
+            this.bindHelper('liveOffline' + this.bindPostfix, function () {
+	            this.bindOnceHelper('liveOnline' + this.bindPostfix, function () {
```
OrenMe, line 184: Is there a chance that we get liveOffline couple of times and then each time do a bind once?

yairans: for the same offline scenario? no.

OrenMe: Probably yes, wondering if we need a state flag as well but it gets ugly cause we also have it in hlsjs as well.

PR #4154, opened 2020-05-18
feat: send error report to kava

OrenMe:
* add key field to error event to match what kanalony is looking for, leave code for backward compatibility
* add error details to kanalony report
* add hlsjs failed url in case of level or fragment error
* add response headers, if applicable, to error object

### DIFF
--- a/modules/Dash/resources/dash.js
+++ b/modules/Dash/resources/dash.js
@@ -521,7 +521,7 @@ try {
 				};
 				if (error.category) {
 					errorObj.code = error.category + "000";
-					errorObj.key = error.category + "000";
+					errorObj.key = errorObj.code";
 				}
 				this.getPlayer().triggerHelper('embedPlayerError', errorObj);
 			mw.log("Dash::Error: ", error);
--- a/modules/Dash/resources/dash.js
+++ b/modules/Dash/resources/dash.js
@@ -521,7 +521,7 @@ try {
 				};
 				if (error.category) {
 					errorObj.code = error.category + "000";
-					errorObj.key = errorObj.code";
+					errorObj.key = errorObj.code;
 				}
 				this.getPlayer().triggerHelper('embedPlayerError', errorObj);
 			mw.log("Dash::Error: ", error);
--- a/modules/Dash/resources/dash.js
+++ b/modules/Dash/resources/dash.js
@@ -521,6 +521,7 @@ try {
 				};
 				if (error.category) {
 					errorObj.code = error.category + "000";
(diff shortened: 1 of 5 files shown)


modules/KalturaSupport/resources/mw.kAnalony.js
```diff
@@ -400,7 +400,8 @@
 
 			this.embedPlayer.bindHelper('playerError', function (e, errorObj) {
                 var errorCode = errorObj && errorObj.key;
```
yairans, line 402: easier to change this to:
```suggestion
                var errorCode = errorObj && errorObj.code;
```
no?

modules/Dash/resources/dash.js
```diff
@@ -521,6 +521,7 @@ try {
 				};
 				if (error.category) {
 					errorObj.code = error.category + "000";
+					errorObj.key = error.category + "000";
```
yairans, line 524:
```suggestion
					errorObj.key = errorObj.code";
```
but see my Q in kAnalony.js

modules/KalturaSupport/resources/mw.kAnalony.js
OrenMe: as mentioned in description - trying to keep backward compatibility and didn't want to go everywhere where we set  key and change to code as it might be used by external dependencies

modules/Dash/resources/dash.js
OrenMe: yes, same end result and there's no logic here.

modules/Dash/resources/dash.js
```diff
@@ -521,6 +521,7 @@ try {
 				};
 				if (error.category) {
 					errorObj.code = error.category + "000";
+					errorObj.key = errorObj.code";
```
yairans, line 524:
```suggestion
					errorObj.key = errorObj.code;
```

modules/Hlsjs/resources/hlsjs.js
```diff
@@ -544,22 +544,35 @@
...
-							networkDetails: data.networkDetails
+							networkDetails: data.networkDetails,
+							headers: headers
 						};
 						var errorObj = {
 							message: JSON.stringify(dataObj),
 							// hls fatal error code could be either Network Error (1000) or Media Errors (3000)
-							code: data.type === "networkError" ? "1000" : "3000"
+							code: data.type === "networkError" ? "1000" : "3000",
+							key: data.type === "networkError" ? "1000" : "3000"
```
yairans, line 569: use computed

PR #4160, opened 2020-05-27
PSVAMB-13273 fix for getting qna container when offline (fix undefined messages)

adimachmali:


GElkayam:
@adimachmali @noam-arad , when is this expected to be merged?

PR #4163, opened 2020-06-02
fix(SUP-22436): playlist thumb are not loading when loadThumbnailWithKs is set

shalommeoded:
The issue is that thumbnailUrl is set to the first parameter that is declared, just changeing the order resolves this. 

Solves SUP-22436

### DIFF
--- a/modules/KalturaSupport/components/chapters/chapters.js
+++ b/modules/KalturaSupport/components/chapters/chapters.js
@@ -481,9 +481,7 @@
 				var customData = item.partnerData ? JSON.parse(item.partnerData) : {};
 				var title = item.title || customData.title;
 				var description = item.description || customData.desc;
-				var thumbnailUrl = mw.getConfig('loadThumbnailWithKs')
-					? _this.getThumbUrl(item.thumbnailUrl || customData.thumbUrl || item)
-					: item.thumbnailUrl || customData.thumbUrl || _this.getThumbUrl(item);
+				var thumbnailUrl = item.thumbnailUrl || customData.thumbUrl || _this.getThumbUrl(item);
 				var thumbnailRotatorUrl = _this.getConfig('thumbnailRotator') ? _this.getThumRotatorUrl() : '';
 				mediaItem = {
 					order: orderId++,
--- a/modules/KalturaSupport/components/chapters/chapters.js
+++ b/modules/KalturaSupport/components/chapters/chapters.js
@@ -481,7 +481,9 @@
 				var customData = item.partnerData ? JSON.parse(item.partnerData) : {};
 				var title = item.title || customData.title;
 				var description = item.description || customData.desc;
-				var thumbnailUrl = item.thumbnailUrl || customData.thumbUrl || _this.getThumbUrl(item);
+				var thumbnailUrl = mw.getConfig('loadThumbnailWithKs')
+					? _this.getThumbUrl(item.thumbnailUrl || customData.thumbUrl || item)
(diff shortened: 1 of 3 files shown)


modules/KalturaSupport/components/playlistAPI.js
```diff
@@ -397,7 +397,7 @@
 				title = kWidget.sanitize( title );
 				description = kWidget.sanitize( description );
 
-				var thumbnailUrl = item.thumbnailUrl || customData.thumbUrl || this.getThumbUrl(item);
+				var thumbnailUrl = this.getThumbUrl(item) || item.thumbnailUrl || customData.thumbUrl;
```
OrenMe, line 400: there is a reason for this order, you want it to take it from item or custom level before falling back to default.

OrenMe: If issue is KS we need to enforce the `loadThumbnailWithKs` settings option here as well.

OrenMe: or maybe handle the item.thumbnailUrl and customData.thumbUrl in `this.getThumbUrl` call

modules/KalturaSupport/components/playlistAPI.js
```diff
@@ -397,15 +397,15 @@
 				title = kWidget.sanitize( title );
 				description = kWidget.sanitize( description );
 
-				var thumbnailUrl = this.getThumbUrl(item) || item.thumbnailUrl || customData.thumbUrl;
+				var thumbnailUrl = item.thumbnailUrl || customData.thumbUrl || this.getThumbUrl(item);
```
OrenMe: same for custom data, so you can make this generic.

shalommeoded: Can you spesify?

OrenMe: `customData.thumbUrl`

modules/KalturaSupport/components/chapters/chapters.js
```diff
@@ -481,7 +481,9 @@
 				var customData = item.partnerData ? JSON.parse(item.partnerData) : {};
 				var title = item.title || customData.title;
 				var description = item.description || customData.desc;
-				var thumbnailUrl = item.thumbnailUrl || customData.thumbUrl || _this.getThumbUrl(item);
+				var thumbnailUrl = mw.getConfig('loadThumbnailWithKs')
```
OrenMe, line 484: this is incorrect! you are now basically making the item.thumbnailUrl and customData.thumbUrl obsolete if the flag is set.
You need to take the respective url and add to it a ks if flag is set and not use the `getThumbUrl` global url settings.

shalommeoded: Done, also no need to add this logic to the chapters plugin, KS is already added.

modules/KalturaSupport/components/playlistAPI.js
```diff
@@ -397,7 +397,7 @@
 				title = kWidget.sanitize( title );
 				description = kWidget.sanitize( description );
 
-				var thumbnailUrl = item.thumbnailUrl || customData.thumbUrl || this.getThumbUrl(item);
+				var thumbnailUrl = this.getThumbUrl(item, customData.thumbUrl);
```
OrenMe, line 400: if you are passing the custom url then in KBaseMediaList.js you need to change the if condition for the custom asset!

PR #4164, opened 2020-06-08
feat(FEC-10004): remove statistics plugin from default

OrenMe:
keep it commented for ansible script that might need to add it for onPrem

### DIFF
--- a/modules/KalturaSupport/UiConfResult.php
+++ b/modules/KalturaSupport/UiConfResult.php
@@ -209,7 +209,7 @@ class UiConfResult {
 
 		// Add Core plugins
 		$basePlugins = array(
-//			'statistics' => array(),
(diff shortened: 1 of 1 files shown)


modules/KalturaSupport/UiConfResult.php
```diff
@@ -209,7 +209,7 @@ public function parseJSON( $uiConf ) {
 
 		// Add Core plugins
 		$basePlugins = array(
-			'statistics' => array(),
+//			'statistics' => array(),
```
GElkayam, line 212: @OrenMe, it would be helpful if the comment marks will be right before the name of the plugin.

OrenMe: done

PR #4166, opened 2020-06-14
fix(SUP-20488): info screen is broken when adding a link to the description

shalommeoded:
Removed the title attribute from the temp file, and added it later on the bind. 

Solves SUP-20488


modules/KalturaSupport/components/info/info.tmpl.html
```diff
@@ -2,6 +2,6 @@
 	<div class='title' title="{mediaProxy.entry.name}" tabindex="0">{mediaProxy.entry.name}</div>
 	<span class="icon-infoScreen-clock" ></span><span class='created' title="Uploaded on {mediaProxy.entry.createdAt|dateFormat}" tabindex="0" >Uploaded on {mediaProxy.entry.createdAt|dateFormat}</span>
     <span class="icon-infoScreen-eye"></span><span class='plays' title="{mediaProxy.entry.plays|numberWithCommas} plays" tabindex="0">{mediaProxy.entry.plays|numberWithCommas} plays</span>
-	<div class='description' title="{mediaProxy.entry.description}" tabindex="-1">{mediaProxy.entry.description}</div>
+	<div class='description' tabindex="-1">{mediaProxy.entry.description}</div>
```
OrenMe, line 5: why do we even put title on this element?
Even for accessibility there are other ways to make screen readers read content of a div and not describe it by setting a title.
@eitanavgil FYI

PR #4167, opened 2020-06-14
fix(SUP-22588): Off button is not customisable

shalommeoded:


modules/KalturaSupport/resources/mw.ClosedCaptions.js
```diff
@@ -1216,7 +1216,7 @@
 		addOffButton: function() {
 			var _this = this;
 			this.getMenu().addItem({
-				'label': 'Off',
+				'label': gM('mwe-off-button'),
```
OrenMe, line 1219:
```suggestion
				'label': gM('mwe-closedCaptions-off-button'),
```

PR #4178, opened 2020-08-02
Toggle off polls when poll change is just a local change (Webcast hid…

adimachmali:
…e polls)

### DIFF
--- a/includes/DefaultSettings.php
+++ b/includes/DefaultSettings.php
@@ -20,7 +20,7 @@ $wgScriptCacheDirectory = realpath( dirname( __FILE__ ) ) . '/cache';
 $wgBaseMwEmbedPath = realpath( dirname( __FILE__ ) . '/../' );
 
 // The version of the library:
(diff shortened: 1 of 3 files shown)


includes/DefaultSettings.php
```diff
@@ -20,7 +20,7 @@
 $wgBaseMwEmbedPath = realpath( dirname( __FILE__ ) . '/../' );
 
 // The version of the library:
-$wgMwEmbedVersion = '2.83.1';
+$wgMwEmbedVersion = '2.83.2';
```
OrenMe, line 23: @adimachmali do not change this file manually, we have automated release process for this

adimachmali: @OrenMe Pls note that the only commit I added yesterday it this:
https://github.com/kaltura/mwEmbed/pull/4178/commits/1edcd95878e89e138e24af6cdd19953a1fe8bcf9
which is not including the bump.

the other files are from previous PRs.

PR #4185, opened 2020-08-16
feat(FEV-715): notify of showing and hiding the QNA UI

eitanavgil:


modules/QnA/resources/qna.js
```diff
@@ -158,8 +157,10 @@
 					_this.getQnaContainer();
 					if (openQnaContainer){
 						qnaObject.show();
+						_this.getPlayer().sendNotification("qna", {visible:true} );
 					} else {
 						qnaObject.hide();
+						_this.getPlayer().sendNotification("qna", {visible:false} );
```
adimachmali, line 163: should you add this also to line: 515 when:
_this.getQnaContainer().find(".qnaModuleBackground").hide();

PR #4189, opened 2020-09-29
fix(SUP-24104): Auto play ignores bumper

shalommeoded:
Set the vid.volume to 0 if we are on autoPlay


OrenMe (commented):
maybe see if we can take the main player volume instead

shalommeoded:
@OrenMe 
Won't work since for some reason the player volume is still set to 0, however, the vid.volume is still set to 1.

OrenMe:
@shalommeoded but this is exactly what I was referring to - take the volume from player always, don't check autoplay - and I would assume we do the exact check in player as well - so avoid duplicating the logic all around.

OrenMe (approved):

kaltura-hooks:
Hi @shalommeoded,

PR #4201, opened 2021-01-29
Added video resolutions in source selector

petur03:
Added some additional flavors to the source selector, accounting for videos that are of greater resolution than 1080P.


kaltura-hooks:
Hi @petur03,

PR #4220, opened 2021-05-05
fix: quiz Korean translation

yairans:
fix bad space format and remove leading and trailing spaces


OrenMe:
Remove trailing spaces as well.
Did we verify this actually cause the &nbsp string to appear?

OrenMe (approved):

PR #4237, opened 2021-06-30
fix(FEC-11373): blank space in QnA area when onPage is false

lianbenjamin:
**issue:**
when qna.onPage=false and the qna is enabled via the webcasting app, there is an empty black container, instead of the qna layout. The reason for this behavior is that the display att of the relevant element stays with "none" value, and never being changed to "block".

**solution:**
adding call to show() on the relevant element where needed.

Solves FEC-11373


yairans (approved):

yairans:
@lianbenjamin please update the base to 2.90

PR #4240, opened 2021-07-30
fix(SUP-28319): prevent focus chapters when side panel closed

semarche-kaltura:
1) use css property "visibility: hidden" to prevent tabbing into closed sidebar panel;
2) handle "home"\"end" keys for navigation to first\last chapter(slide)

### DIFF
--- a/modules/KalturaSupport/components/chapters/chapters.js
+++ b/modules/KalturaSupport/components/chapters/chapters.js
@@ -1294,6 +1294,19 @@
 					}
 				}
 			});
+			this.handleHomeEndKeys = function (e) {
+				if(e.keyCode === 35) {
+					// end key pressed
+					e.stopPropagation();
+					this.getMediaListDomElements().filter( ".chapterBox" ).last().focus();
+				}
+				if(e.keyCode === 36) {
+					// home key pressed
+					e.stopPropagation();
+					this.getMediaListDomElements().filter( ".chapterBox" ).first().focus();
+				}
+			};
+
 			this.onItemKey = function (e) {
 				var _this = this;
                 if(e.keyCode === 13){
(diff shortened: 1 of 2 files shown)


modules/KalturaSupport/components/chapters/chapters.js
```diff
@@ -1311,10 +1324,12 @@
...
 			var mediaBoxes = this.getMediaListDomElements();
-			mediaBoxes.on('mousedown mouseup mouseout', function(){
-				this.blur();
-			})
-			.on('keyup', $.proxy(this.onItemKey , this));
+			mediaBoxes
+				.on('mousedown mouseup mouseout', function(){
+					this.blur();
+				})
+				.on('keydown', $.proxy(this.handleHomeEndKeys, this))
```
OrenMe, line 1331: why separate keydown and keyup? does it matter if we use any of them? if not lets consolidate to one and reuse `onItemKey` for all

semarche-kaltura: player handle 'keyup' before plugin and makes seekToBeginning and seekToEnd

OrenMe: For this case there is `onDisableKeyboardBinding` event to disable the main player keyboard bindings and `onEnableKeyboardBinding` to bring them back.

RoyBregman: I think for this scenario, Yuval created this:
https://github.com/kaltura/playkit-js-ui/blob/master/docs/with-keyboard-event.md
@OrenMe @yairans ?

semarche-kaltura: cool,
I'll make the changes with `onDisableKeyboardBinding` and `onEnableKeyboardBinding`

RoyBregman: As the solution @OrenMe suggested turns out to be much more complex (due to other `onEnableKeyboardBinding` and `onDisableKeyboardBinding` which are fired irrelevantly) we will still stick with @semarche-kaltura solution

RoyBregman (approved):

PR #4250, opened 2021-09-14
fix(FEC-11542): tooltips in full screen don't pop out when hovering over it

lianbenjamin:
**the issue:**
when in fullscreen, tooltips of bottom bar buttons are not displayed.

**solution:**
increase z-index of tooltips to 9999, as player ins fullscreen is having z-index=9999.
also had to delay the focus of play/pause btn when entering fullscreen, to prevent from the tooltip to appear in a wrong place.

Solves FEC-11542


kaltura-hooks:
Hi @lianbenjamin,

yairans (approved):

PR #4253, opened 2021-12-08
fix(FEC-11747): v2 aria label of quiz previous question btn is wrong

lianbenjamin:
**the issue:**
When the user presses the tab key or down arrow to navigate to the previous question button, the aria label actually says "Previous Question button - disabled". This makes it sound like screen reader users do not have the option to navigate to the previous question (even though they do).

**root cause:**
"mwe-quiz-previous-question" has value "Previous Question button - disabled".
Note that there is a dedicated key for when it's actually disabled- "mwe-quiz-previous-question-disabled", which has the same value.

**solution:**
remove word "disabled" from string value of key "mwe-quiz-previous-question".

Solves FEC-11747


yairans (approved):

yairans (changes requested):
@lianbenjamin please open 2.92 and merge to it

yairans (approved):

PR #4255, opened 2021-12-22
fix(FEC-11791): share url has cross-site scripting vulnerability

lianbenjamin:
**description of the issue:**
when share plugin is enabled for the player, the share url is not being sanitized, which exposes a security vulnerability.

**the solution:**
sanitizing encoded and decoded share url.

Solves FEC-11791

### DIFF
--- a/modules/KalturaSupport/components/share/share.js
+++ b/modules/KalturaSupport/components/share/share.js
@@ -616,7 +616,8 @@
 					res = document.URL;
 				}
 			}
-			return kWidget.sanitize(decodeURIComponent(res));
+			var sanitized = kWidget.sanitize(decodeURIComponent(res));
+			return encodeURIComponent(sanitized);
 		}
 		// -------------- finish setup player url according to the socialShareURL flashvar ------- //
 
--- a/modules/KalturaSupport/components/share/share.js
+++ b/modules/KalturaSupport/components/share/share.js
@@ -616,7 +616,7 @@
 					res = document.URL;
 				}
 			}
-			return res;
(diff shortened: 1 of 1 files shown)


modules/KalturaSupport/components/share/share.js
```diff
@@ -616,7 +616,7 @@
 					res = document.URL;
 				}
 			}
-			return res;
+			return kWidget.sanitize(decodeURIComponent(res));
```
yairans, line 619: do we need to encode the url again?

lianbenjamin: no, the decodeURIComponent replaces each escape sequence in the encoded with the character that it represents in ascii. if it's not an escaping char, it moves on. for example:
https://www.kaltura.com/.../embed/dynamic?%22%3E%3Cimg%20src=x%20onerror=prompt(4)%3E
would turn into:
https://www.kaltura.com/.../embed/dynamic?">
by doing so, the sanitizer would be able to replace any escaping chars with empty strings. without that decoding stage, sanitizer won't work in cases where there is encoded uri. if the uri is not encoded (most cases), the decoder does nothing and returns the uri string as-is.

yairans:
@OrenMe please take a look on this PR

modules/KalturaSupport/components/share/share.js
yairans: i think encode it again it safer. if it was encoded we have to keep it encoded. if not we just encode the url, what the side effect? @OrenMe wdyt?

yairans (approved):

PR #4256, opened 2022-01-09
fix(FEC-11831): embedded captions in dual screen are shown when autoPlay is true

lianbenjamin:
**the issue:**
when autoplay=true and secondary video has embedded captions, the embedded captions are still shown on player.

**solution:**
hide embedded captions from secondary screen when loadedmetadata event is triggered.

Solves FEC-11831

### DIFF
--- a/modules/KalturaSupport/components/dualScreen/dualScreen.js
+++ b/modules/KalturaSupport/components/dualScreen/dualScreen.js
@@ -269,17 +269,6 @@
 				this.bind( 'onplay', function () {
 						if (!isOverlayScreenOpen && _this.canManipulateControlViews()) {
 							_this.controlBar.enable();
-							// Added this in order to remove the Embedded Captions that are showing on the second screen
-							// As the HLSjs is only disabling it for the main player
-							if (!_this.embedPlayer.getKalturaConfig('closedCaptions', 'showEmbeddedCaptions')) {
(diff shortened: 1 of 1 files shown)


modules/KalturaSupport/components/dualScreen/dualScreen.js
```diff
@@ -405,6 +413,7 @@
 
                         var _this = this;
                         this.initView();
+						this.hideEmbeddedCaptionsFromSecondary();
```
yairans, line 416: fix the indentation. 
also, if you move it after `initControlBar` so you can keep the logic like was before by using `canManipulateControlViews`

yairans (changes requested):

modules/KalturaSupport/components/dualScreen/dualScreen.js
lianbenjamin: do you think we should still use canManipulateControlViews function?.. how hiding the embedded captions is related to controlBar or if the view mode is locked?

yairans: I said you can not you should. if you want to make the change as small as possible

lianbenjamin: @yairans i opened the github editor and the indentation doesn't exist there, nor in my webstorm... seems to be only here.

yairans (approved):

PR #4259, opened 2022-03-02
fix(FEC-12016): poll responses is showing 0 after toggling between the 'Hide/Show' button within the Webcasting App

lianbenjamin:
**the issue:**
the scenario:
1. presenter is creating poll
2. users are voting
3. presenter clicks on hide via webcasting app
4. presenter clicks on show via webcasting app

expected behavior -> player displays the correct amount of responses (totalVoters)
observed behavior -> player is displaying 0 responses

**root cause:**
player is not saving poll-results cuepoints, which contains the total voters value. since hide is resetting the local data, when clicking on show, the player is displaying 0 responses.

**the solution:**
when player is receiving poll-results cuepoint, we should save it.

Solves FEC-12016

### DIFF
--- a/modules/WebcastPolls/resources/webcastPolls.js
+++ b/modules/WebcastPolls/resources/webcastPolls.js
@@ -283,10 +283,6 @@
 						            _this.pollData.pollResults.totalVoters = cuepointContent.totalVoters;
 						            _this.view.syncDOMPollResults();
 					            }
-                                // kCuePoint filter is sorting according to startTime - cuepoint has start time 0
-                                cuepoint.startTime = cuepoint.createdAt * 1000;
-                                // need to save poll-results cuepoints in case we reset internal data
-                                _this.getPlayer().kCuePoints.updateCuePoints([cuepoint]);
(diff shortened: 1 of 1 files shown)


modules/WebcastPolls/resources/webcastPolls.js
```diff
@@ -283,7 +283,11 @@
 						            _this.pollData.pollResults.totalVoters = cuepointContent.totalVoters;
 						            _this.view.syncDOMPollResults();
 					            }
-                                // make sure we process poll-results cuepoint 
+                                // kCuePoint filter is sorting according to startTime - cuepoint has start time 0
+                                cuepoint.startTime = cuepoint.createdAt * 1000;
+                                // need to save poll-results cuepoints in case we reset internal data
+                                _this.getPlayer().kCuePoints.updateCuePoints([cuepoint]);
```
yairans, line 289: What's about doing it once after for?

yairans: just the update action

yairans (approved):

PR #4264, opened 2022-04-20
fix(FEC-12183): chapter and slides have wrong thumbnails

lianbenjamin:
**the issue:**
chapters show wrong thumbnail. when clicking on a thumbnail, we jump to a point in the video where the thumbnail is different from the chapter's thumbnail.

**root cause:**
player is doing parseInt to time/1000, therefore, the time is less accurate (i.e. 118.123 becomes 118). in V7 we send time/1000 w/o parsing it to int or changing the value.

**solution:**
remove parseInt and leave the time value as accurate as it can be.

Solves FEC-12183

### DIFF
--- a/modules/KalturaSupport/resources/mw.KBaseMediaList.js
+++ b/modules/KalturaSupport/resources/mw.KBaseMediaList.js
@@ -426,7 +426,7 @@
 			var thumbUrl = kWidgetSupport.getKalturaThumbnailUrl(
 				$.extend( {}, this.baseThumbSettings, {
                     'url': this.embedPlayer.evaluate('{mediaProxy.entry.thumbnailUrl}'),
-					'vid_sec': parseInt( time / 1000 )
+					'vid_sec': time / 1000
 				} )
 			);
 			return thumbUrl;


modules/KalturaSupport/resources/mw.KBaseMediaList.js
```diff
@@ -426,7 +426,7 @@
 			var thumbUrl = kWidgetSupport.getKalturaThumbnailUrl(
 				$.extend( {}, this.baseThumbSettings, {
                     'url': this.embedPlayer.evaluate('{mediaProxy.entry.thumbnailUrl}'),
-					'vid_sec': parseInt( time / 1000 )
```
yairans, line 429: @lianbenjamin did you check why the parsing was added?

lianbenjamin: @yairans yes, tried to search for a reason using git blame but didn't find anything specific to parseInt. maybe BE in the past didn't support float number...

yairans (approved):

PR #4284, opened 2022-06-13
fix(FEC-12322): add alternatives to getReferer logic

lianbenjamin:
add alternatives to getReferer logic.

**the issue:**
in some cases, OSs/browsers, for example, OS X/Safari, will not pass $_SERVER['HTTP_REFERER'] and then the fallback would be 'http://www.kaltura.com/'. If a customer is using an ACP where certain domains are allowed (excluding kaltura.com) and the browser did not pass the HTTP_REFERER, then the player will display an error.

**solution:**
add alternatives to get/create the referer.

Solves FEC-12322

### DIFF
--- a/modules/KalturaSupport/RequestHelper.php
+++ b/modules/KalturaSupport/RequestHelper.php
@@ -209,13 +209,33 @@ class RequestHelper {
 		if( $wgKalturaForceReferer !== false ){
 			return $wgKalturaForceReferer;
 		}
-		if( isset( $_SERVER['HTTP_REFERER'] ) ){
-			$urlParts = parse_url( $_SERVER['HTTP_REFERER'] );
-			if (isset( $urlParts['scheme'] ) &&  isset( $urlParts['host']) ) {
-				return $urlParts['scheme'] . "://" . $urlParts['host'] . "/";
-			}
+		if (isset( $_SERVER['HTTP_REFERER'] )){
+            $refererUrl = $this->buildReferer($_SERVER['HTTP_REFERER']);
+        } else if (isset( $_SERVER ) && isset($_SERVER['HTTP_HOST'])) {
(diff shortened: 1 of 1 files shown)


modules/KalturaSupport/RequestHelper.php
```diff
@@ -209,15 +209,35 @@ public function getReferer(){
...
+			$refererUrl = $this->buildReferer($_SERVER['HTTP_REFERER']);
+		} else if (isset( $_SERVER ) && isset($_SERVER['HTTP_HOST'])) {
+			if ( isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] === 'on') {
+				$refererUrl = "https://";
+			} else {
+				$refererUrl = "http://";
 			}
+			$refererUrl.= $_SERVER['HTTP_HOST'];
+		} else {
+			$requestHeaders = getallheaders();
```
OrenMe, line 222: Isn't this the same as server http_referrer? I think internally php takes it from the header.

lianbenjamin: the root cause here is that sometimes http_referer does not exist at all (although http_host does). since i think we can't know for 100% which other use cases this change might affect, I preferred to keep the old original logic and to have it as first option, to avoid regressions. the goal was to "expand" and cover more cases. does that make sense?

OrenMe: I'm referring only to first vs third options, I think server http_referrer BW get all headers.referrer are the same so this is redundant.
Also, will http host ever be empty?

lianbenjamin: checking online the request headers vs server, seems they are the same like you mentioned - removing the 3rd option as it is the same as the 1st, thanks.
according to php manual the http_host is header from the current request, **if there is one**. I am not sure if it means that it can be empty or just not exist.. but why not using the empty? as it covers both cases anyway

giladna (approved):

PR #4298, opened 2022-08-31
feat(FEC-12521): fallback option upon inaccessible API Gateway

lianbenjamin:
**context:**
adding a fallback option upon inaccessible API Gateway.
BE are adding new property to `KalturaAccessControlModifyRequestHostRegexAction` named `checkAliveTimeoutMs` - see ref [here](https://www.kaltura.com/api_v3/testmeDoc/objects/KalturaAccessControlModifyRequestHostRegexAction.html), which defines the timeout duration for checking the ip.

**solution:**
- when the new property `checkAliveTimeoutMs` is defined and its value is bigger than 0, an ajax call to the serviceUrl is being made.
- the URLs replacement will occur only if the connection succeeded or `checkAliveTimeoutMs` is 0, null, negative number.
- in case `checkAliveTimeoutMs` is a valid value (bigger than 0) and the ajax call has failed, we are not overriding the urls.

**description of changes:**
- add async ajax call with timeout
- use promise on `updatePlayerContextData` to prevent async issues

Solves FEC-12521

### DIFF
--- a/modules/KalturaSupport/resources/mw.KWidgetSupport.js
+++ b/modules/KalturaSupport/resources/mw.KWidgetSupport.js
@@ -329,7 +329,12 @@ mw.KWidgetSupport.prototype = {
 							}
 						}
 					}
+				} else {
+					deferred.resolve();
 				}
+			}
+			else {
+				deferred.resolve();
 			}
 		}
 		return deferred.promise();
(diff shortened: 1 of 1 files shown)


GElkayam:
Hi @lianbenjamin , FYI, the [server side PR](https://github.com/kaltura/server/pull/11791/files) was merged and is planned to be deployed to SaaS with Rigel-18.14.0

SivanA-Kaltura (approved):

giladna (approved):

PR #4301, opened 2022-09-28
fix(FEC-12600): player should support referer without scheme part

lianbenjamin:
**the issue:**
sometimes referer header might not include "http" or "https" (i.e. kaltura.com). in this case, if kaltura.com is an allowed domain in an ACP, it will not be recognized by the player, since the player is looking for "scheme" part (http/s) and "host" part. we are using an php function called parse_url to break the referer header into url parts. in cases where the referer does not contain a scheme part (http/s), there is also no host, but "path" exists instead.

**solution:**
when building the referer, adding another check for "path" attribute.

Solves FEC-12600


giladna:
@OrenMe 
could you please look at this PR and put your concernes if any regarding the fix 
as usually valid referrer should come with schema and here we try to change the behaviour

giladna (approved):

PR #4309, opened 2023-05-16
fix(SUP-37385): user cannot see q&a

inbalvasserman:
the issue:
user cannot see q&a.

solution:
in case onPage= false: add show() for qna 
in case onPage=True: add the elements for qna, and change display:block to the button

### DIFF
--- a/modules/QnA/resources/qna.js
+++ b/modules/QnA/resources/qna.js
@@ -150,6 +150,7 @@
 							_this.getPlayer().toggleFullscreen() ;
 							openQnaContainer=true;
 						}
+						_this.getQnaContainer();
 						if (openQnaContainer){
 							qnaObject.show();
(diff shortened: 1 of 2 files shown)


modules/QnA/resources/qna.js
```diff
@@ -138,38 +138,33 @@
...
-                onVideoTogglePluginButton.on("click", function(){
-
-                    var openQnaContainer=!qnaObject.is(":visible");
-
-                    if (_this.getPlayer().layoutBuilder.fullScreenManager.isInFullScreen()) {
-                        _this.getPlayer().toggleFullscreen() ;
-                        openQnaContainer=true;
-                    }
-
-					_this.getQnaContainer();
```
giladna, line 157: why this was removed?

inbalvasserman: fixed

giladna (approved):

PR #4312, opened 2023-08-16
fix(SUP-38476): [EU Cloud] - issues with video broadcast

inbalvasserman:
ISSUE:
when use eCDN with player in pop-up window, when open different player the url changing 

SOLUTION:
save the original url in window.parent.mv so when open new window the variable will still save

### DIFF
--- a/modules/KalturaSupport/resources/mw.KWidgetSupport.js
+++ b/modules/KalturaSupport/resources/mw.KWidgetSupport.js
@@ -20,7 +20,6 @@ mw.KWidgetSupport.prototype = {
 	kClient : null,
 	kSessionId: null, // Used for Analytics events
 	originalStreamerType: null,
-	originalServiceUrl: null,
 
 	// Constructor check settings etc
 	init: function( options ){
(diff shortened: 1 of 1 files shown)


modules/KalturaSupport/resources/mw.KWidgetSupport.js
```diff
@@ -302,10 +301,12 @@ mw.KWidgetSupport.prototype = {
 
 					if (action.pattern && action.replacement) {
 						var regExp=new RegExp(action.pattern, "i");
-						if (!this.originalServiceUrl){
-							this.originalServiceUrl = mw.config.get('Kaltura.playManifestServiceUrl');
+						var originalServiceUrl = window.parent.mw.getConfig('Kaltura.originalServiceUrl');
```
lianbenjamin, line 304: more elegant-
`var originalServiceUrl = window.parent.mw.getConfig('Kaltura.originalServiceUrl') || mw.config.get('Kaltura.playManifestServiceUrl');`

modules/KalturaSupport/resources/mw.KWidgetSupport.js
```diff
@@ -302,10 +301,12 @@ mw.KWidgetSupport.prototype = {
...
+						if (!originalServiceUrl){
+							originalServiceUrl = mw.config.get('Kaltura.playManifestServiceUrl');
+							window.parent.mw.setConfig('Kaltura.originalServiceUrl', originalServiceUrl)
```
lianbenjamin, line 307: also, is it possible that `originalServiceUrl` is `null` or `undefined`?

lianbenjamin (changes requested):

modules/KalturaSupport/resources/mw.KWidgetSupport.js
inbalvasserman: it not the same. the propose is to save the originalServiceUrl because the playManifestServiceUrl is override. the bug was that it used to take the playManifestServiceUrl after it was set and it cause to use the wrong url and so the player didnt play

modules/KalturaSupport/resources/mw.KWidgetSupport.js
inbalvasserman: yes

lianbenjamin (approved):

PR #4314, opened 2024-06-07
fix(ONPREM-38): php-8 fixes for legacy v2 player

navarro967:
This PR introduces changes to the v2 player to allow it to run in php 8 environments. It also includes changes from v2.101 - v2.103.


navarro967:
@inbalvasserman @JonathanTGold @yossipapi

